PR #8, opened 2019-11-21
Feature/web basics

ch-rfishwick:
Created minimal springboot application


.idea/sonarIssues.xml
```diff
@@ -38,6 +58,46 @@
             <set />
           </value>
         </entry>
+        <entry key="$PROJECT_DIR$/../../company-accounts.web.ch.gov.uk/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/ApprovalController.java">
```
Ngriffiths1: I think these entries should be deleted as they reference company accounts? Also I think this file should be in the gitignore?

PR #16, opened 2019-12-02
Feature/fa273 change to controllers

Ngriffiths1:
Changed controllers to reflect naming conventions of paths and files. Created home controller and corresponding html file.


src/main/java/uk/gov/ch/developer/docs/controller/developer/GettingStartedController.java
```diff
@@ -6,25 +6,25 @@
...
+@RequestMapping("/dev-hub/getting-started")
 public class GettingStartedController extends BaseController {
 
-    public static final String TEMPLATE_NAME = "docs/gettingStarted";
+    private static final String DEV_HUB_GETTING_STARTED = "dev-hub/gettingStarted";
     private static final String TEMPLATE_TITLE = "Getting started with the Companies House API";
 
     @GetMapping
-    public String get() {
+    public String getDevHubHome() {
```
ch-dclarke: Get() is sufficient as we are on the specific controller already

PR #23, opened 2019-12-06
Fixed link to main content

ch-rfishwick:
Fixed tab link to main page content

__JIRA Ticket Number__
FA-276

### Type of change

*  Bug fix
* [ ] New feature
* [ ] Breaking change

### Pull request checklist


src/main/resources/templates/layouts/baseLayout.html
```diff
@@ -11,7 +11,9 @@
       rel="stylesheet"
       th:href="@{{cdnUrl}/stylesheets/chs-styles.css(cdnUrl=${@environment.getProperty('cdn.url')})}"
       type="text/css"/>
-
+  <link
+      href="http://chs-dev.internal:8080/stylesheets/assets-digital-cabinet-office-gov-uk-static/govuk-template.css"
```
CH-JonathanBlake: This doesn't seem right at all.
We can't hard code this to the development server.

src/main/resources/templates/fragments/header.html
```diff
@@ -5,7 +5,6 @@
 </head>
 <body class="govuk-template__body">
 <th:block th:fragment="header">
-  <a class="govuk-skip-link" href="#main-content">Skip to main content</a>
```
CH-JonathanBlake: https://www.w3.org/TR/WCAG20-TECHS/G1.html
This is reccommended by W3C, if we're removing it then we ought to replace it with something

src/main/resources/templates/layouts/baseLayout.html
Ngriffiths1: same

src/main/resources/templates/fragments/header.html
Ngriffiths1: same

ch-rfishwick: I did, it's in the baseLayout

CH-JonathanBlake: Do you mean that hard coded thing that is pointing at a .css?

ch-rfishwick: No, the other change to the base layout

CH-JonathanBlake: Sorry, I forgot that it wasn't the only change. Deleted the other comment

src/main/resources/templates/layouts/baseLayout.html
```diff
@@ -29,6 +31,11 @@
 <script
     th:src="@{{cdnUrl}/javascripts/vendor/jquery-3.3.1.min.js(cdnUrl=${@environment.getProperty('cdn.url')})}"></script>
 <div class="entire-wrapper">
+  <div id="skiplink-container">
+    <div>
```
ch-dclarke: Why is there a <div> inside an <div> here?

ch-rfishwick: Because that is how it is written on the current site. But it can be removed

ch-dclarke: Please remove it if it serves no purpose.

PR #37, opened 2020-01-13
Feature/session service and web security

ch-rfishwick:
Addition of SessionService class

__JIRA Ticket Number__
FA-332

### Type of change

* [ ] Bug fix
* New feature
* [ ] Breaking change

### Pull request checklist


CH-JonathanBlake:
> Where is this used?

Was that part of the acceptance criteria? It will be used in the user interceptor but I don't think that can be implemented before the security adds the filters, because that is where sessions are created. And Zain picked up the security? But Interceptor will be added in Userbar shows username.

ch-rfishwick:
> Where is this used?


> > Where is this used?
> 
> Was that part of the acceptance criteria? It will be used in the user interceptor but I don't think that can be implemented before the security adds the filters, because that is where sessions are created. And Zain picked up the security? But Interceptor will be added in Userbar shows username.

It will be added in the user interceptor which we will implement with FA-392. And yes, Zain has picked up security.

PR #38, opened 2020-01-14
Feature/fa 344 navbar user version

CH-JonathanBlake:
Altered navigation bar fragments and model to change their behaviour based on parameters of the user model.
__JIRA Ticket Number__
FA-344
### Type of change

* [ ] Bug fix
* [X] New feature
* [X] Breaking change

### Pull request checklist


src/main/java/uk/gov/ch/developer/docs/config/NavBarConfig.java
```diff
@@ -25,25 +25,24 @@
     public NavBarModel getNavBarModel() {
 
         NavBarModel model = new NavBarModel();
-        NavItemList manageApplications = model.addHeading("Manage Applications");
+        NavItemList manageApplications = model
+                .addHeading("Manage Applications", UserRequired.userRequired);
```
CH-JonathanBlake: I created this enum because I thought it was ambiguous if true meant it checked for users or if it required users.
Interested if others think it was helpful of just unnecessary.

src/main/java/uk/gov/ch/developer/docs/models/nav/NavBarItem.java
```diff
@@ -4,56 +4,116 @@
 import java.util.Collections;
 import java.util.List;
 
+/**
+ * Basic implementation of the INavBarItem.
+ */
 public class NavBarItem implements INavBarItem {
 
+    private final String DEFAULTURL = System.getProperties().getProperty("home.url");
```
CH-JonathanBlake: Sonar seems to want this camel cased however I don't know that, that would be typical of what is basically a constant. Opinions?

src/main/java/uk/gov/ch/developer/docs/config/NavBarConfig.java
ch-dclarke: That's sensible.

src/main/java/uk/gov/ch/developer/docs/models/nav/NavBarItem.java
ch-dclarke: it's reasonable of Sonar, as it is a non-constant function result.

src/main/resources/templates/fragments/navBarSubMenu.html
```diff
@@ -6,18 +6,21 @@
 </head>
 <body>
 
+<!-- This is expected to be called in the context of a list -->
 <div th:fragment="navBarSubMenu (menu)" th:remove="tag">
-  <!-- This is expected to be called in the context of a list -->
-  <li class="app-subnav__section-item">
+  <th:block
```
ch-dclarke: This brings the business logic into the view. That's really a dangerous design practice.

It would be better if the model understands the condition and then returns only viewable items.

On any given page display, the menu won't change.

CH-JonathanBlake:
Closing pull request as this has been moved out of sprint.

PR #44, opened 2020-01-17
Feature/fa 392 display user details

ch-rfishwick:
Addition of user details interceptor and appropriate unit tests. The interceptor has been added to the main application, but its addition to the interceptor registry has been temporarily commented out until a functional sign-in process exists, since its implementation currently leads to server error.

__JIRA Ticket Number__
FA-392

### Type of change

* [ ] Bug fix
* New feature
* [ ] Breaking change

### Pull request checklist

* I have added unit tests for new code that I have added


src/main/java/uk/gov/ch/developer/docs/DocsWebApplication.java
```diff
@@ -14,10 +15,13 @@
...
 
     private LoggingInterceptor loggingInterceptor;
+    private UserDetailsInterceptor userDetailsInterceptor;
 
     @Autowired
-    public DocsWebApplication(LoggingInterceptor loggingInterceptor) {
+    public DocsWebApplication(LoggingInterceptor loggingInterceptor,
+            UserDetailsInterceptor userDetailsInterceptor) {
         this.loggingInterceptor = loggingInterceptor;
+        this.userDetailsInterceptor = userDetailsInterceptor;
```
Zain-Abbas1: Lets comment this line out as well until we sort out the sign-in work.

src/main/java/uk/gov/ch/developer/docs/interceptor/LoggingInterceptor.java
```diff
@@ -15,12 +15,11 @@
 @Component
 public class LoggingInterceptor extends HandlerInterceptorAdapter {
 
+    private final Logger logger;
     private RequestLogger requestLogger = new RequestLogger() {
         // Implement Defaults
```
Zain-Abbas1: Delete comment

src/main/resources/templates/fragments/header.html
```diff
@@ -165,4 +165,4 @@
   </header>
 </th:block>
 </body>
-</html>
+</html>
```
Zain-Abbas1: Were the above changes done by you? Seems to be a lot of unrelated changes, is this branch up to date with master?

src/main/java/uk/gov/ch/developer/docs/interceptor/UserDetailsInterceptor.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
+import org.springframework.web.servlet.view.UrlBasedViewResolver;
+import uk.gov.ch.developer.docs.session.SessionService;
+
+@Component
+public class UserDetailsInterceptor extends HandlerInterceptorAdapter {
+
+    private static final String USER_EMAIL = "userEmail";
+    private static final String SIGN_IN_KEY = "signin_info";
+    private static final String USER_PROFILE_KEY = "user_profile";
```
ch-dclarke: Should these keys be private, as I would guess the same constant values need to be used as keys elsewhere?

src/main/java/uk/gov/ch/developer/docs/interceptor/UserDetailsInterceptor.java
```diff
@@ -0,0 +1,44 @@
...
+                && !modelAndView.getViewName()
+                .startsWith(UrlBasedViewResolver.REDIRECT_URL_PREFIX)))) {
+
+            Map<String, Object> sessionData = sessionService.getSessionDataFromContext();
+            Map<String, Object> signInInfo = (Map<String, Object>) sessionData.get(SIGN_IN_KEY);
+            if (signInInfo != null) {
+                Map<String, Object> userProfile = (Map<String, Object>) signInInfo
+                        .get(USER_PROFILE_KEY);
+                modelAndView.addObject(USER_EMAIL, userProfile.get(EMAIL_KEY));
+            }
```
ch-dclarke: Aren't these logically better handled through the methods on the `HttpServletRequest`, such as:

`String getRemoteUser()`
Returns the login of the user making this request, if the user has been authenticated, or null if the user has not been authenticated. Whether the user name is sent with each subsequent request depends on the browser and type of authentication. Same as the value of the CGI variable REMOTE_USER.
Returns:
a String specifying the login of the user making this request, or null if the user login is not known

`java.security.Principal getUserPrincipal()`
Returns a java.security.Principal object containing the name of the current authenticated user. If the user has not been authenticated, the method returns null.
Returns:
a java.security.Principal containing the name of the user making this request; null if the user has not been authenticated

src/main/resources/templates/fragments/header.html
ch-rfishwick: I did a reverse merge with master and then reformatted before pushing. I'm a bit confused because any changes that have crept in from the master merge shouldn't be causing disparities with master. The reformatting may have changed some....

src/main/java/uk/gov/ch/developer/docs/interceptor/LoggingInterceptor.java
CH-JonathanBlake: I disagree. I think this is worth putting here to say that we are deliberately not implementing or overriding methods.

src/main/java/uk/gov/ch/developer/docs/interceptor/UserDetailsInterceptor.java
```diff
@@ -0,0 +1,44 @@
...
+    public void postHandle(HttpServletRequest request, HttpServletResponse response, Object handler,
+            @Nullable ModelAndView modelAndView) throws Exception {
+
+        if (modelAndView != null && modelAndView.getViewName() != null
+                && (request.getMethod().equalsIgnoreCase("GET")
+                || (request.getMethod().equalsIgnoreCase("POST")
+                && !modelAndView.getViewName()
+                .startsWith(UrlBasedViewResolver.REDIRECT_URL_PREFIX)))) {
+
+            Map<String, Object> sessionData = sessionService.getSessionDataFromContext();
```
CH-JonathanBlake: When me and David designed the User Model we extended it from LFPWeb. I think we were expecting the interceptor to have a (an autowired?) UserModel field and that would be populated running the populateUserDetails method against the user and then adding that to the model.
This way we don't have to change the User Interceptor if we need to update the user model

src/main/java/uk/gov/ch/developer/docs/interceptor/LoggingInterceptor.java
Zain-Abbas1: Why would that be of any benefit? As a default non-descriptive comments should be removed or expanded on.

src/main/java/uk/gov/ch/developer/docs/interceptor/UserDetailsInterceptor.java
CH-JonathanBlake: Since this doesn't really have anything to do with security I don't really mind.
User Profile and SignInInfo here are ch classes not standard Java so I'm not sure they have those methods.

PR #45, opened 2020-01-22
Feature/fa 392 retrieve user details

CH-JonathanBlake:
Altered controller behaviour to gather user model. Login/out functionality is still dependent on the filters being delivered.
__JIRA Ticket Number__
fa-392
### Type of change

* [ ] Bug fix
* [X] New feature
* [X] Breaking change

### Pull request checklist


src/main/resources/templates/fragments/userBar.html
```diff
@@ -9,8 +9,8 @@
     <ul class="mobile-hidden" id="navigation">
       <li>
         <ul class="mobile-hidden"
-            th:if="${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(user)}">
-          <li class="user" id="signed-in-user" th:text="${user.getId() + ':'}"></li>
+            th:if="${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}">
```
ch-dclarke: Is it necessary to have the full class name, rather than the shortened one?

src/main/java/uk/gov/ch/developer/docs/controller/ModelAttributeNames.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.ch.developer.docs.controller;
+
+public class ModelAttributeNames {
```
ch-dclarke: Make `final static`

src/main/resources/templates/fragments/userBar.html
CH-JonathanBlake: Spoken in person, but yes, this needs to be fully type qualified. If we want to change this we can either refactor UserModel or change the default spring type locator in the expression language.
Either choice is more work than we care to do atm.

pom.xml
```diff
@@ -7,6 +7,45 @@
...
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <environmentVariables>
+            <CACHE_POOL_SIZE>10</CACHE_POOL_SIZE>
+            <CACHE_SERVER>localhost:1234</CACHE_SERVER>
+            <COOKIE_DOMAIN>.companieshouse.gov.uk</COOKIE_DOMAIN>
+            <COOKIE_NAME>abc</COOKIE_NAME>
+            <COOKIE_SECRET>iamasecret</COOKIE_SECRET>
+            <DEFAULT_SESSION_EXPIRATION>3600</DEFAULT_SESSION_EXPIRATION>
+            <HUMAN_LOG>1</HUMAN_LOG>
```
Zain-Abbas1: Are the values correct? Doesn't seem like something we would put into live?

pom.xml
```diff
@@ -7,6 +7,45 @@
...
+
+            <account.url>//</account.url>
+
+            <createAccount.url>${home.url}/create-account</createAccount.url>
+            <developerGuidelines.url>${home.url}/developer-guidelines</developerGuidelines.url>
+            <gettingStarted.url>${home.url}/getting-started</gettingStarted.url>
+            <manageApplications.url>${home.url}/manage-applications</manageApplications.url>
+            <home.url>${home.url}</home.url>
+            <piwik.siteId>//</piwik.siteId>
+            <piwik.url>//</piwik.url>
```
Zain-Abbas1: Why have we put the environment variable in the pom.xml when we have the applications.properties file?

ch-dclarke: The values for surefire only apply in unit tests.

pom.xml
ch-dclarke: The Surefire plugin only applies to unit tests. They are will not be live values.

CH-JonathanBlake: The values were take directly from the session handler as well.

Zain-Abbas1: 👍

pom.xml
```diff
@@ -133,6 +172,10 @@
     <thymeleaf-layout-dialect.version>2.3.0</thymeleaf-layout-dialect.version>
 
     <web-security-java.version>1.1.0-rc1</web-security-java.version>
+
+
+    <home.url></home.url>
```
Zain-Abbas1: Should this be where the other env variables are declared?

pom.xml
Zain-Abbas1: 👍

Zain-Abbas1:
Sonar has failed, needs looking at.

pom.xml
CH-JonathanBlake: This is declared here so that it can be referenced by the other Environment variables, like <gettingStarted.url>${home.url}/getting-started</gettingStarted.url>.
I think normally we would prefer to not do this in the pom generally because they aren't contained.

CH-JonathanBlake:
> Sonar has failed, needs looking at.

Its upset because we didn't get tests to cover the dummy controllers and we're printing the stack trace error of an error that should never be possible via System.out.println.

Since they're temporary code that hopefully will be removed by the end of the next sprint I don't think anyone considers it a priority.

pom.xml
Zain-Abbas1: I see, it is a bit confusing as it's not contained. However we have the same line on line 32, which looks to be performing the same function.

CH-JonathanBlake: No. This line is declaring home.url to be a POM property, line 32 is saying Set Enviroment Variable 'home.url' to be the same as POM property 'home.url'
I do agree that its a bit confusing though

Zain-Abbas1: Right makes sense, I'll approve it and maybe later on we can see if there is a less confusing way of implementing it.

PR #47, opened 2020-01-24
Feature/fa 210 update static pages

CH-JonathanBlake:
Responding to Fungai's testing comments.
__JIRA Ticket Number__
FA-210
### Type of change

* [X] Bug fix
* [ ] New feature
* [ ] Breaking change

### Pull request checklist
No unit testable code was added.


CH-JonathanBlake:
> Please check through the classes for the heading types and use correct classes. Also might be worth a check for any classes that can be added to "section" as well. Other than minor changes all looks good.

Not entirely certain what you mean by heading types, I think you're complaining about the use of the small heading class on some h3's, would be good to confirm this though.
Sections have no classes as far as I can tell in the gds specification and since they're purely semantic I wouldn't expect them to have classes really. The current live site uses classes sections for the anchor links.

PR #51, opened 2020-02-05
CH-MojoX-SignIn - Implement sign in functionality

Zain-Abbas1:
Initial stab at porting the signin functionality from `CH-MojoX-SignIn` to java. This is being done inline currently with a view to breaking this out into a separate library in the near future.
 - Signin controller - redirects user to accounts authentication page
- UserCallbackController - Will handle the callback from the server in future tickets - for the time being it is being hit and the Code/State values are being logged.
- Dummy controllers have been deleted as that was only a temporary fix, the callback work will implement the real functionality.

__JIRA Ticket Number__
FA-389

### Type of change

* [ ] Bug fix
* [ ] New feature
* [ ] Breaking change

### Pull request checklist


src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
```diff
@@ -0,0 +1,231 @@
...
+        if (queryString != null)
+            originalRequestUrl.append("?").append(queryString);
+
+        // Set the scope
+        String scope = null;
+        if (companyNumber != null)
+            scope = OAUTH_COMPANY_SCOPE_PREFIX + companyNumber;
+
+        // Generate and store a nonce in the session
+        Session sessionToUpdate = session;
```
CH-JonathanBlake: Why a new variable? Why do we get username from the older one? Can't we just use the one that was passed in?

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
```diff
@@ -0,0 +1,231 @@
...
+     * 
+     * @param session User session from which to retrieve the email address
+     * @return email
+     */
+    private String getEmailFromSession(Session session) {
+
+        String email = "";
+
+        SignInInfo signInInfo = new SignInInfo();
+        if (session != null) {
```
CH-JonathanBlake: If session is null can't we just return null or an empty string? Because we know that the new SignInInfo has no user profile or content?

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
```diff
@@ -0,0 +1,231 @@
...
+     * @param scope Scope of the request
+     * @param nonce
+     * @param email Email address from the session, empty if not present
+     * @return Authorisation URI
+     */
+    protected String createAuthoriseURIWithForceAndHint(String originalRequestUri, String scope,
+            String nonce, String email) {
+
+        StringBuilder sb = new StringBuilder(createAuthoriseURI(originalRequestUri, scope, nonce));
+        sb.append("&reauthenticate=force");
```
CH-JonathanBlake: Stringbuilder append returns itself so we could chain these methods.

src/main/java/uk/gov/ch/developer/docs/security/WebSecurity.java
```diff
@@ -0,0 +1,29 @@
...
+            http.antMatcher("/signin");
+        }
+    }
+
+    @Configuration
+    @Order(2)
+    public static class DocsWebSecurityFilterConfig extends WebSecurityConfigurerAdapter {
+        @Override
+        protected void configure(HttpSecurity http) throws Exception {
+            http.antMatcher("/**");
```
CH-JonathanBlake: I still think things should be secure by default.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -0,0 +1,60 @@
...
+import uk.gov.companieshouse.environment.EnvironmentReader;
+import uk.gov.companieshouse.environment.impl.EnvironmentReaderImpl;
+import uk.gov.companieshouse.logging.Logger;
+import uk.gov.companieshouse.logging.LoggerFactory;
+
+@Controller
+@RequestMapping("${callback.url}")
+public class UserCallbackController {
+
+    protected Logger LOGGER = LoggerFactory.getLogger("docs.developer.ch.gov.uk");
```
CH-JonathanBlake: Same comment about Logger being protected not private?

CH-JonathanBlake:
What did you do to allow it to merge without any review approval?

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
```diff
@@ -0,0 +1,231 @@
...
+
+    private static final EnvironmentReader reader = new EnvironmentReaderImpl();
+
+    private static final String base64Key = reader.getMandatoryString("OAUTH2_REQUEST_KEY");
+    private static final String authorizationUri = reader.getMandatoryString("OAUTH2_AUTH_URI");
+    private static final String clientId = reader.getMandatoryString("OAUTH2_CLIENT_ID");
+    private static final String redirectUri = reader.getMandatoryString("OAUTH2_REDIRECT_URI");
+    private static final String secret = reader.getMandatoryString("COOKIE_SECRET");
+
+    protected Logger LOGGER = LoggerFactory.getLogger("docs.developer.ch.gov.uk");
```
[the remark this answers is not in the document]
ch-dclarke: Loggers should normally be `private final static` as theirs context should be tied quite tightly to the specific class and allows for class level changes of logging detail.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
```diff
@@ -0,0 +1,231 @@
...
+     * @throws IOException
+     */
+    protected void redirectForAuth(Session session, HttpServletRequest request,
+            HttpServletResponse response, String companyNumber, boolean force) throws IOException {
+
+        // Find the original requested url
+        StringBuilder originalRequestUrl = new StringBuilder(request.getRequestURL().toString());
+        String queryString = request.getQueryString();
+        if (queryString != null)
+            originalRequestUrl.append("?").append(queryString);
```
ch-dclarke: Coding standards - always use braces for `if` statements, so that code is consistent and bugs aren't introduced by insertion of additional code. 

Please use the code formatting plugin to enforce CH standards.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
```diff
@@ -0,0 +1,231 @@
...
+
+        byte[] key = Base64.decodeBase64(base64Key);
+
+        try {
+            jweObject.encrypt(new DirectEncrypter(key));
+        } catch (JOSEException e) {
+            LOGGER.error(e, null);
+        }
+
+        return jweObject.serialize();
```
ch-dclarke: Is it risky to return the object if there was an exception thrown one the encryption?

dalebradley:
Unit tests

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
```diff
@@ -0,0 +1,231 @@
...
+    protected Logger LOGGER = LoggerFactory.getLogger("docs.developer.ch.gov.uk");
+
+    @GetMapping
+    public void getSignIn(ServletRequest request, ServletResponse response) throws IOException {
+
+        Session chSession =
+                (Session) request.getAttribute(SessionHandler.CHS_SESSION_REQUEST_ATT_KEY);
+
+        HttpServletRequest httpServletRequest = (HttpServletRequest) request;
+        HttpServletResponse httpServletResponse = (HttpServletResponse) response;
```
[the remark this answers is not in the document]
Zain-Abbas1: Yeah I will change this now, this was initially done as part of a filter which takes the ServletRequest etc as parameters.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
```diff
@@ -0,0 +1,231 @@
...
+    private static final String base64Key = reader.getMandatoryString("OAUTH2_REQUEST_KEY");
+    private static final String authorizationUri = reader.getMandatoryString("OAUTH2_AUTH_URI");
+    private static final String clientId = reader.getMandatoryString("OAUTH2_CLIENT_ID");
+    private static final String redirectUri = reader.getMandatoryString("OAUTH2_REDIRECT_URI");
+    private static final String secret = reader.getMandatoryString("COOKIE_SECRET");
+
+    protected Logger LOGGER = LoggerFactory.getLogger("docs.developer.ch.gov.uk");
+
+    @GetMapping
+    public void getSignIn(ServletRequest request, ServletResponse response) throws IOException {
```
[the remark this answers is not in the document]
Zain-Abbas1: How do you mean?

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
Zain-Abbas1: This is carrying existing functionality from the filters, not quite sure why it was done this way. Perhaps an extra layer of security, protecting against possible hijacking? As the nonce will be stored in this new session and not the one retrieved from context.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
Zain-Abbas1: We could, I see no benefit other than syntactical?

src/main/java/uk/gov/ch/developer/docs/security/WebSecurity.java
Zain-Abbas1: This to be dealt with in a separate ticket, but at the moment we are leaving it like this until I am sure the functionality works.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
```diff
@@ -0,0 +1,231 @@
...
+import uk.gov.companieshouse.session.model.UserProfile;
+
+@Controller
+@RequestMapping("/signin")
+public class SignInController {
+
+    private static final String OAUTH_COMPANY_SCOPE_PREFIX =
+            "https://api.companieshouse.gov.uk/company/";
+
+    private static final EnvironmentReader reader = new EnvironmentReaderImpl();
```
ch-dclarke: This probably should be an injected implementation.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
Zain-Abbas1: During the encryption or during the serialisation? As during the encryption an exception would be caught and logged out.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
Zain-Abbas1: As part of the application.properties file?

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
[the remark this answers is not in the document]
Zain-Abbas1: Done

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
ch-dclarke: During the encryption, the exception is caught and logged, but the code then continues to serialise the object for return. I would be much happier if, in the `catch` the code returned an empty result, or an error result, rather than dropping through to the `return jweObject.serialize` in case of potential exposure of elements that should be encrypted.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
ch-dclarke: I was referring to the EnvironmentReader instance. I would expect the implementation to be a bean, rather than explicitly created. This gives us better options for testing and differing implementations if necessary for different environments.

CH-JonathanBlake: @Autowired basically.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
CH-JonathanBlake: Isn't that a worthwhile benefit if changes need to happen anyway?

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
ch-dclarke: Similarly, where a `new SessionImpl` is performed lower in the code, it ought to be obtained from the factory, so that we can again mock it and replace it easily, although that won't be auto-wired, I suspect.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
CH-JonathanBlake: Zain I know you marked this as resolved but I don't actually see the braces that David wanted.

Zain-Abbas1: Don't know what happened there, I'll push the change up again.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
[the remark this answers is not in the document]
CH-JonathanBlake: I was incorrect here, because we aren't pulling parameters out of a request and are asking for the whole request we shouldn't use the annotation.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
Zain-Abbas1: SessionFactory doesn't have an implementation to return a brand new session directly.

Zain-Abbas1: Implemented the java bean for the env reader

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
Zain-Abbas1: I have returned an empty string after logging the exception.

CH-JonathanBlake: Oh no! David hates multiple exit points.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private static final String base64Key = reader.getMandatoryString("OAUTH2_REQUEST_KEY");
+
+    @GetMapping
+    @ResponseBody
+    public String getParams(@RequestParam("state") String state, @RequestParam("code") String code)
+            throws ParseException, KeyLengthException, JOSEException,
+            net.minidev.json.parser.ParseException {
+        LOGGER.info("State: " + state);
+        LOGGER.info("Code: " + code);
```
ch-dclarke: I think these should be logged at Debug level, if they are logged at all.
Is there any possibility of revealing security information in the logs?

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
```diff
@@ -0,0 +1,225 @@
...
+    private String getEmailFromSession(Session session) {
+
+        String email = "";
+
+        SignInInfo signInInfo = new SignInInfo();
+        if (session != null) {
+            signInInfo = session.getSignInInfo();
+        }
+
+        UserProfile userProfile = signInInfo.getUserProfile();
```
ch-dclarke: Can you successfully obtain the email address with this code?

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
```diff
@@ -0,0 +1,225 @@
...
+import net.minidev.json.JSONObject;
+import uk.gov.companieshouse.environment.EnvironmentReader;
+import uk.gov.companieshouse.logging.Logger;
+import uk.gov.companieshouse.logging.LoggerFactory;
+import uk.gov.companieshouse.session.Session;
+import uk.gov.companieshouse.session.SessionImpl;
+import uk.gov.companieshouse.session.SessionKeys;
+import uk.gov.companieshouse.session.handler.SessionHandler;
+import uk.gov.companieshouse.session.model.SignInInfo;
+import uk.gov.companieshouse.session.model.UserProfile;
```
ch-dclarke: I'm concerned that we have so many imports. 
Is it possible that we are trying to do too much in this one class. Could we break some of the functionality into another class, to separate concerns and consider _Single Responsibility_ issues?

ch-dclarke: Maybe separating out the JSON and JWE handling would be logical.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
Zain-Abbas1: Correct - these have been removed

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
Zain-Abbas1: Not yet - part of future work

PR #52, opened 2020-02-17
Rewrite callback function in Java from CH-MojoX-SignIn

Zain-Abbas1:
- Add user callback controller to handle logic for callback - this includes the decryption of state and checking of nonce values
- Add web security class including session filters for each route
- Add a new Bean `WebServerFactoryCustomizer` which allows a `.`domain 

NOTE: This is currently the first part of the callback, the rest will be covered in future tickets.

__JIRA Ticket Number__
FA-494

### Type of change

* [ ] Bug fix
* [ ] New feature
* [ ] Breaking change

### Pull request checklist


src/main/java/uk/gov/ch/developer/docs/DocsWebApplication.java
```diff
@@ -28,4 +32,15 @@ public static void main(String[] args) {
     public void addInterceptors(InterceptorRegistry registry) {
         registry.addInterceptor(loggingInterceptor);
     }
+    
+    @Bean
+    public WebServerFactoryCustomizer<TomcatServletWebServerFactory> webServerFactoryCustomizer() {
```
CH-JonathanBlake: As raised in offline discussion, this may not be needed.
If it is, could we have it documented why we wanted to override our servers default cookie processor?

src/main/java/uk/gov/ch/developer/docs/security/WebSecurity.java
```diff
@@ -5,25 +5,39 @@
...
+
+    @Configuration
+    @Order(3)
+    public static class CallbackSecurityConfig extends WebSecurityConfigurerAdapter {
         @Override
         protected void configure(HttpSecurity http) throws Exception {
-            http.antMatcher("/**");
+            http.antMatcher("/oauth2/user/**").addFilterBefore(new SessionHandler(),
+                    BasicAuthenticationFilter.class);
         }
```
ch-dclarke: We should probably combine these into one `@Configuration` as we have the same handler for all of the matched paths.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -30,17 +33,34 @@
...
+        Session sessionCallback = sessionService.getSessionFromContext();
+
+        Map<String, Object> sessionMap = sessionCallback.getData();
+        String nonceSession = sessionMap.get(SessionKeys.NONCE.getKey()).toString();
+
+        String nonceState = decryptState(state);
 
-        sessionService.getSessionDataFromContext();
+        if (nonceState.equals(nonceSession)) {
+            LOGGER.info("Nonce values are equal");
```
ch-dclarke: This should probably log at Debug, rather an Info level

src/main/java/uk/gov/ch/developer/docs/security/WebSecurity.java
CH-JonathanBlake: Please post words

PR #54, opened 2020-02-25
Feature/fa 496 identity provider

ch-dclarke:
Identity provider and OAuth specific code added
Checking of Nonce performed.

N.B. No unit tests

__JIRA Ticket Number__
FA-496

### Type of change

* [ ] Bug fix
* New feature
* [ ] Breaking change

### Pull request checklist

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,5 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="EntryPointsManager">
+    <list size="1">
+      <item index="0" class="java.lang.String" itemvalue="org.springframework.stereotype.Controller" />
+    </list>
+  </component>
   <component name="MavenProjectsManager">
     <option name="originalFiles">
       <list>
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,15 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="EntryPointsManager">
-    <list size="3">
-      <item index="0" class="java.lang.String" itemvalue="org.springframework.stereotype.Component" />
-      <item index="1" class="java.lang.String" itemvalue="org.springframework.stereotype.Controller" />
-      <item index="2" class="java.lang.String" itemvalue="org.springframework.web.bind.annotation.GetMapping" />
-    </list>
-    <writeAnnotations>
-      <writeAnnotation name="org.springframework.beans.factory.annotation.Autowired" />
-    </writeAnnotations>
-  </component>
   <component name="MavenProjectsManager">
     <option name="originalFiles">
       <list>
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,9 +1,14 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="EntryPointsManager">
-    <list size="1">
-      <item index="0" class="java.lang.String" itemvalue="org.springframework.stereotype.Controller" />
+    <list size="3">
+      <item index="0" class="java.lang.String" itemvalue="org.springframework.stereotype.Component" />
+      <item index="1" class="java.lang.String" itemvalue="org.springframework.stereotype.Controller" />
+      <item index="2" class="java.lang.String" itemvalue="org.springframework.web.bind.annotation.GetMapping" />
     </list>
+    <writeAnnotations>
+      <writeAnnotation name="org.springframework.beans.factory.annotation.Autowired" />
+    </writeAnnotations>
   </component>
   <component name="MavenProjectsManager">
     <option name="originalFiles">
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
@@ -78,6 +78,30 @@ public class SignInController {
         response.sendRedirect(authoriseUri);
     }
 
+    /**
+     * Constructs the authorisation URI with additional force and hint
+     *
+     * @param originalRequestUri Original URI from which to redirect
+     * @param scope Scope of the request
+     * @param email Email address from the session, empty if not present
+     * @return Authorisation URI
+     */
+    protected String createAuthoriseURIWithForceAndHint(final String originalRequestUri,
+            final String scope, final String nonce, final String email) {
+        if (email == null) {
+            LOGGER.debug("No email supplied");
+        }
+        if (email == null) {
+            LOGGER.debug("No scope supplied");
+        }
+        final String hint = oauth.oauth2EncodeState(email, nonce, "email");
+        final String authUrl = identityProvider.getAuthorisationUrl(originalRequestUri, scope);
+        String authUri = authUrl + "&reauthenticate=force"
+                + "&hint="
+                + hint;
+        return authUri;
+    }
+
     //TODO Move this onto the OAuth instance
     private String generateSessionNonce(final Session session) {
         // Generate and store a nonce in the session
@@ -96,7 +120,7 @@ public class SignInController {
      * @param session User session from which to retrieve the email address
      * @return email
      */
-    private String getEmailFromSession(Session session) {
+    private String getEmailFromSession(final Session session) {
 
         String email = "";
 
@@ -104,7 +128,7 @@ public class SignInController {
         if (session != null) {
             signInInfo = session.getSignInInfo();
         }
-        UserProfile userProfile = signInInfo.getUserProfile();
+        final UserProfile userProfile = signInInfo.getUserProfile();
         if (userProfile != null) {
             email = userProfile.getEmail();
         }
@@ -112,6 +136,7 @@ public class SignInController {
     }
 
     //TODO Move this into the OAuth object
+
     /**
      * Generates a secure unique key
      *
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
@@ -78,6 +78,7 @@ public class SignInController {
         response.sendRedirect(authoriseUri);
     }
 
+    //TODO Move this onto the OAuth instance
     private String generateSessionNonce(final Session session) {
         // Generate and store a nonce in the session
         Session sessionToUpdate = session;
@@ -110,6 +111,7 @@ public class SignInController {
         return email;
     }
 
+    //TODO Move this into the OAuth object
     /**
      * Generates a secure unique key
      *
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
@@ -9,19 +9,13 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
-import com.nimbusds.jose.EncryptionMethod;
-import com.nimbusds.jose.JOSEException;
-import com.nimbusds.jose.JWEAlgorithm;
-import com.nimbusds.jose.JWEHeader;
-import com.nimbusds.jose.JWEObject;
-import com.nimbusds.jose.Payload;
-import com.nimbusds.jose.crypto.DirectEncrypter;
-import net.minidev.json.JSONObject;
+import uk.gov.ch.developer.docs.oauth.IIdentityProvider;
+import uk.gov.ch.developer.docs.oauth.IOauth;
+import uk.gov.ch.developer.docs.session.SessionFactory;
 import uk.gov.companieshouse.environment.EnvironmentReader;
 import uk.gov.companieshouse.logging.Logger;
 import uk.gov.companieshouse.logging.LoggerFactory;
 import uk.gov.companieshouse.session.Session;
-import uk.gov.companieshouse.session.SessionImpl;
 import uk.gov.companieshouse.session.SessionKeys;
 import uk.gov.companieshouse.session.handler.SessionHandler;
 import uk.gov.companieshouse.session.model.SignInInfo;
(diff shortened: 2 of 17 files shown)


src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -1,76 +1,48 @@
...
+            @RequestParam("code") String code) {
+        LOGGER.trace("Code:" + code);
+        LOGGER.trace("State:" + state);
+
+        final String returnedNonce = getNonceFromState(state);
+        if (!oauth.oauth2VerifyNonce(returnedNonce)) {
+            LOGGER.error("Invalid nonce value in state during oauth2 callback");
+            return "redirect:/";
+        } else {
+            return ("redirect:" + identityProvider.getRedirectUri());
```
Zain-Abbas1: This is incorrect, we don't want to redirect back to the callback once nonce values have been asserted to be equal. In face we could probably do away with the else statement as this redirect should be the last redirect after all the logic of callback has been implemented. At the moment I have redirected it back to home page, there is a ticket in the backlog to look at proper redirection back to the page where the user has initiated the sign-in request from.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
```diff
@@ -9,107 +9,90 @@
...
+        final String nonce = generateSessionNonce(session);
 
-        // Set the scope
-        String scope = null;
-        if (companyNumber != null) {
-            scope = OAUTH_COMPANY_SCOPE_PREFIX + companyNumber;
-        }
+        // Build oauth uri and redirect
+        final String state = oauth.oauth2EncodeState(originalRequestUrl, nonce, "content");
+        final String authoriseUri = identityProvider.getAuthorisationUrl(state);
```
Zain-Abbas1: I've noticed the `createAuthoriseURIWithForceAndHint` method has been removed. From my last conversations with @macreman this was decided to be left in for future-proofing as this was related to protecting the session from being hijacked etc.

.idea/misc.xml
```diff
@@ -1,5 +1,15 @@
...
+  <component name="EntryPointsManager">
+    <list size="3">
+      <item index="0" class="java.lang.String" itemvalue="org.springframework.stereotype.Component" />
+      <item index="1" class="java.lang.String" itemvalue="org.springframework.stereotype.Controller" />
+      <item index="2" class="java.lang.String" itemvalue="org.springframework.web.bind.annotation.GetMapping" />
+    </list>
+    <writeAnnotations>
+      <writeAnnotation name="org.springframework.beans.factory.annotation.Autowired" />
+    </writeAnnotations>
+  </component>
```
Zain-Abbas1: This looks like a file that is generated by IntelliJ, do we need this?

Zain-Abbas1:
General comment, this current version of identity providers is different to what we have seen with other implementations i.e., with the use of a .yml or .properties file. It looks as though we are just reading in environment values and if this is the way we want to go I would suggest looking into how the current developer site has implemented it's identity providers, ensuring we are matching the same values which will prove useful for further iterations. It seems as though we are missing important parameters such as the token url etc, which is important for future iterations.

src/main/java/uk/gov/ch/oauth/IdentityProvider.java
```diff
@@ -0,0 +1,84 @@
...
+    final private String clientSecret;
+
+    @Autowired
+    public IdentityProvider(final EnvironmentReader reader) {
+        requestKey = decoder.decode(reader.getMandatoryString("DEVELOPER_OAUTH2_REQUEST_KEY"));
+        this.authorizationUri = reader.getMandatoryString("OAUTH2_AUTH_URI");
+        this.clientId = reader.getMandatoryString("CHS_DEVELOPER_CLIENT_ID");
+        redirectUri = reader.getMandatoryString("OAUTH2_REDIRECT_URI");
+        clientSecret = reader.getMandatoryString("CHS_DEVELOPER_CLIENT_SECRET");
+       // TODO is this better? redirectUri = reader.getMandatoryString("REDIRECT_URI");
```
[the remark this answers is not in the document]
Zain-Abbas1: Just to add to the first comment, I've left a comment under where you've done the redirect and why I think it's probably not the correct way to do it. So it might be worth keeping the the current version for the time being.

start.sh
```diff
@@ -2,29 +2,30 @@
...
+  wget -O "${APP_DIR}/global_env" "${CONFIG_URL}/${ENVIRONMENT}/global_env"
+  wget -O "${APP_DIR}/app_env" "${CONFIG_URL}/${ENVIRONMENT}/${APP_NAME}/env"
+  source "${APP_DIR}/private_env"
+  source "${APP_DIR}/global_env"
+  source "${APP_DIR}/app_env"
 fi
 
-exec java ${JAVA_MEM_ARGS} -jar -Dserver.port="${PORT}" "${APP_DIR}/docs.developer.ch.gov.uk.jar"
+#exec java ${JAVA_MEM_ARGS} -jar -Dserver.port="${PORT}" "${APP_DIR}/docs.developer.ch.gov.uk.jar"
+exec java ${JAVA_MEM_ARGS} -jar -Dserver.port="${PORT}" -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005 "${APP_DIR}/docs.developer.ch.gov.uk.jar"
```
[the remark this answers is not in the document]
ch-dclarke: Yes, although I'll leave the debugger line commented out.

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -0,0 +1,103 @@
...
+    public Payload oauth2DecodeState(final String state) {
+        Payload payload;
+        try {
+            final JWEObject jweObject = JWEObject.parse(state);
+
+            final byte[] key = identityProvider.getRequestKey();
+            jweObject.decrypt(new DirectDecrypter(key));
+            payload = jweObject.getPayload();
+        } catch (Exception e) {
+            LOGGER.error(e, null);
```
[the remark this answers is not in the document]
CH-JonathanBlake: I'm not sure we can when that exception could be literally anything.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
ch-dclarke: Ah, I was unaware of that conversation. I'll see about resurrecting it.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
ch-dclarke: @Zain-Abbas1 Would you mind correcting the redirect in the next iteration of this code?

CH-JonathanBlake: Would be good to put a TODO there even if you don't change it yourself. I know that I'd forget if I picked this up.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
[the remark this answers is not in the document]
ch-dclarke: Although there is a technical difference between the two, I think it is OK to leave them as they are.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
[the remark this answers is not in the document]
ch-dclarke: These are very distinctly different behaviours. One is the expected happy path, the other is effectively an exception.

ch-dclarke: Yes

src/main/java/uk/gov/ch/developer/docs/session/SessionFactory.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public Session getSessionByCookieId(final String cookieId) {
+        return uk.gov.companieshouse.session.SessionFactory.getSessionByCookieId(cookieId);
+    }
+
+    public Session createSession() {
+        return getSessionByCookieId(null);
+    }
+
+}
```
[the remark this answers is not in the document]
ch-dclarke: It is there to wrap the static SessionHandler in a way that will allow us to write tests that inject Mocks. We can't modify static methods in Mockit 7 Junit.

.idea/misc.xml
ch-dclarke: I agree, but don't know, as it wasn't in the .gitignore exclusions. It's easier to delete and ignore than it is to recreate if necessary.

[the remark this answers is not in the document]
ch-dclarke: It wasn't in the .gitignore

src/main/java/uk/gov/ch/developer/docs/session/SessionFactory.java
```diff
@@ -0,0 +1,33 @@
...
+
+    private SessionFactory() {
+    }
+
+    public SessionImpl getSessionByCookieId(final Store store, final String cookieId) {
+        return uk.gov.companieshouse.session.SessionFactory.getSessionByCookieId(store, cookieId);
+    }
+
+    public Session getSessionByCookieId(final String cookieId) {
+        return uk.gov.companieshouse.session.SessionFactory.getSessionByCookieId(cookieId);
```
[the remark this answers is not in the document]
ch-dclarke: Because there are 2 different implementations of SessionFactory. The one we are in, which wraps the other to enable mocking.

src/main/java/uk/gov/ch/oauth/IdentityProvider.java
```diff
@@ -0,0 +1,84 @@
...
+    @Override
+    public String getClientId() {
+        return clientId;
+    }
+
+    /**
+     * Auth URL with scope added
+     */
+    public String getAuthorisationUrl(final String state, final String scope) {
+        StringBuilder sb = new StringBuilder(getAuthorisationUrl(state));
```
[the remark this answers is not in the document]
ch-dclarke: Correct

src/main/java/uk/gov/ch/oauth/Oauth2.java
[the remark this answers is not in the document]
ch-dclarke: That was my feeling. This is in the category of, this should never fail, but if it does, we need to do log it for diagnosis.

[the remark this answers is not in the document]
ch-dclarke: I've updated the message, as well as possible.

src/main/java/uk/gov/ch/developer/docs/session/SessionFactory.java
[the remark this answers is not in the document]
Zain-Abbas1: Can we remove all unused methods then? By the looks of it we only need `getDefaultStoreImpl()`

src/main/java/uk/gov/ch/oauth/IdentityProvider.java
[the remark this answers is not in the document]
Zain-Abbas1: Can we uncomment this line out please, as it will be used when correcting the redirect.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
Zain-Abbas1: Come to think about it, we could potentially use the Hijack filter which would implement this anyway, so we may not need it after all. I will need to check with @macreman on this.

src/main/java/uk/gov/ch/developer/docs/session/SessionFactory.java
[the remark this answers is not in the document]
Zain-Abbas1: This can be deleted as we are not using this. See my comment above.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
ch-dclarke: If it's not needed, then it should be removed.

src/main/java/uk/gov/ch/developer/docs/session/SessionFactory.java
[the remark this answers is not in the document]
ch-dclarke: This class can't be deleted, if we wish to mock it out.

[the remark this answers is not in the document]
ch-dclarke: I can remove one unused (at present) method.

Zain-Abbas1:
> General comment, this current version of identity providers is different to what we have seen with other implementations i.e., with the use of a .yml or .properties file. It looks as though we are just reading in environment values and if this is the way we want to go I would suggest looking into how the current developer site has implemented it's identity providers, ensuring we are matching the same values which will prove useful for further iterations. It seems as though we are missing important parameters such as the token url etc.

David can you take a look at this (https://github.com/companieshouse/developer.ch.gov.uk/blob/master/appconfig.yml) and include the missing parameters please. Namely - authorize_url, token_url, profile_url.

src/main/java/uk/gov/ch/developer/docs/session/SessionFactory.java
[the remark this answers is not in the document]
ch-dclarke: Done

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -1,76 +1,50 @@
...
-        Session sessionCallback = sessionService.getSessionFromContext();
-
-        Map<String, Object> sessionMap = sessionCallback.getData();
-        String nonceSession = sessionMap.get(SessionKeys.NONCE.getKey()).toString();
-
-        String nonceState = decryptState(state);
-
-        if (nonceState.equals(nonceSession)) {
-            LOGGER.debug("Nonce values are equal");
+    @ResponseBody
```
Zain-Abbas1: We could probably lose this

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -1,76 +1,50 @@
...
+    public String getCallback(@RequestParam("state") String state,
+            @RequestParam("code") String code) {
+        LOGGER.trace("Code:" + code);
+        LOGGER.trace("State:" + state);
```
Zain-Abbas1: We could lose these as well

src/main/java/uk/gov/ch/developer/docs/session/SessionFactory.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private SessionFactory() {
+    }
+
+    public Session getSessionByCookieId(final String cookieId) {
+        return uk.gov.companieshouse.session.SessionFactory.getSessionByCookieId(cookieId);
+    }
+
+    public Session createSession() {
+        return getSessionByCookieId(null);
```
Zain-Abbas1: I agree we should leave this class, can we combine this with above method & change the javadoc.
```suggestion
        return uk.gov.companieshouse.session.SessionFactory.getSessionByCookieId(null);
```

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
```diff
@@ -9,217 +9,143 @@
...
+     * @param scope Scope of the request
+     * @param email Email address from the session, empty if not present
+     * @return Authorisation URI
+     */
+    protected String createAuthoriseURIWithForceAndHint(final String originalRequestUri,
+            final String scope, final String nonce, final String email) {
+        if (email == null) {
+            LOGGER.debug("No email supplied");
+        }
+        if (email == null) {
```
CH-JonathanBlake: Logic check doesn't match Logger message. And is duplicate.

src/main/java/uk/gov/ch/developer/docs/session/SessionFactory.java
ch-dclarke: I would much rather it shadowed the original static class. I see your point, but if we mock out one, I would prefer to keep the same chain of calls.

src/main/java/uk/gov/ch/oauth/IdentityProvider.java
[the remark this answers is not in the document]
ch-dclarke: I wasn't finding the value in my Vagrant environment, but was finding the OAUTH... value, which I suspect should be the correct one, however I noticed @Zain-Abbas1  using the other key in your changes.

src/main/java/uk/gov/ch/oauth/IdentityProvider.java
```diff
@@ -0,0 +1,103 @@
...
+    private String homeUrl;
+
+    @Autowired
+    public IdentityProvider(final EnvironmentReader reader) {
+        requestKey = decoder.decode(reader.getMandatoryString("DEVELOPER_OAUTH2_REQUEST_KEY"));
+        this.authorizationUri = reader.getMandatoryString("OAUTH2_AUTH_URI");
+        this.clientId = reader.getMandatoryString("CHS_DEVELOPER_CLIENT_ID");
+        clientSecret = reader.getMandatoryString("CHS_DEVELOPER_CLIENT_SECRET");
+        redirectUri = reader.getMandatoryString("OAUTH2_REDIRECT_URI");
+        // TODO is this better? Which is the correct source of the redirect URL? redirectUri = reader.getMandatoryString("REDIRECT_URI");
```
CH-JonathanBlake: Do we still want this comment if we've re-implemented the code?

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
ch-dclarke: Yes.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
ch-dclarke: As per our direct conversation, we will leave these in place, for now.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
ch-dclarke: fixed

src/main/java/uk/gov/ch/oauth/IdentityProvider.java
ch-dclarke: This is really a question of where is the source of truth for this value. I could get the OAUTH2... variant from the existing environment.

CH-JonathanBlake: Do we have a plan for resolving the question and an understanding of how to fix this in future?

ch-dclarke: I think a decision needs to be made. My preference would be for an environment variable named so that it's obvious for OAUTH usage.

src/main/java/uk/gov/ch/oauth/IdentityProvider.java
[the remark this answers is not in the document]
Zain-Abbas1: The value should be in your vagrant env, you may need to do a git pull. `OAUTH2_REDIRECT_URI` is the callback redirect, `REDIRECT_URI` is where the app should redirect to once the callback is complete. This is what I mentioned before about me needing to do it this way until the ticket is picked up from the backlog, that will investigate the redirect in a proper way.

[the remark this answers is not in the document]
ch-dclarke: At present, the code works out the redirect for return using the source of the original signing request.

I'm going to have to leave the decision in your hands.

src/main/java/uk/gov/ch/oauth/IdentityProvider.java
Zain-Abbas1: See my comment above, just do a pull on vagrant and you should see this value.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
Zain-Abbas1: I have spoken to Billy and we may need to keep this in after all, I will go over it in detail with him tomorrow.

.idea/misc.xml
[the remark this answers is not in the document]
Zain-Abbas1: I can see .idea in the .gitignore file, I will delete this file.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
Zain-Abbas1: I have added a new redirectUriPage which is just the redirect back to the home page for now.

src/main/java/uk/gov/ch/oauth/IdentityProvider.java
Zain-Abbas1: I have uncommented the line out and changed the var to redirectUriPage - which only deals with the redirect to the home page.

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Extract the OAuth2 Nonce from the current session
+     *
+     * @return The Nonce String from within the session or null if not found.
+     */
+    private String getSessionNonce() {
+        String oauth2Nonce = null;
+        try {
+            final Map<String, Object> data = sessionService.getSessionDataFromContext();
+            oauth2Nonce = (String) data.getOrDefault(".oauth2_nonce", null);
```
Zain-Abbas1: Should use the SessionKeys class instead of `.oauth2_nonce`

Zain-Abbas1: Done

PR #56, opened 2020-03-02
Get OAuth token

Zain-Abbas1:
Add OAuthToken class in order to map the response to the pojo & use new webclient to make post request for oauth token & small change to oauth2 class which allows us to use nonce key from sessionkeys class

__JIRA Ticket Number__
FA-562

### Type of change

* [ ] Bug fix
* [ ] New feature
* [ ] Breaking change

### Pull request checklist


pom.xml
```diff
@@ -159,6 +159,15 @@
...
     </dependency>
+    <dependency>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-starter-webflux</artifactId>
+    </dependency>
+<!--     <dependency>
+        <groupId>org.projectreactor</groupId>
+        <artifactId>reactor-spring</artifactId>
+        <version>1.0.1.RELEASE</version>
+    </dependency> -->
```
CH-JonathanBlake: Can you remove this comment.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -35,6 +40,19 @@ public String getCallback(@RequestParam("state") String state,
...
+        webClient.post().uri(URI.create(identityProvider.getTokenUrl()))
+                .contentType(MediaType.APPLICATION_FORM_URLENCODED)
+                .accept(MediaType.APPLICATION_JSON)
+                .body(BodyInserters.fromPublisher(Mono.just("code=" + code + "&client_id="
+                        + identityProvider.getClientId() + "&client_secret="
+                        + identityProvider.getClientSecret() + "&redirect_uri="
+                        + identityProvider.getRedirectUri() + "&grant_type=authorization_code"),
+                        String.class))
+                .retrieve().bodyToMono(OAuthToken.class)
+                .subscribe(s -> LOGGER.trace("Access Token " + s.getAccess_token()));
```
CH-JonathanBlake: Seems like a bad idea to log access tokens.

src/main/java/uk/gov/ch/developer/docs/controller/developer/OAuthToken.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public String getRefresh_token() {
+        return refresh_token;
+    }
+
+    public void setRefresh_token(String refresh_token) {
+        this.refresh_token = refresh_token;
+    }
+
+    public String getAccess_token() {
```
ch-dclarke: Please name member using Java styling names, i.e. camel case. If they need to be mapped to JSON, then use attributes.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
Zain-Abbas1: This is to prove that we have an access token so it can be tested, it will be removed in the next ticket.

ch-dclarke: Please add a //TODO to remove the logging line later.

src/main/java/uk/gov/ch/developer/docs/controller/developer/OAuthToken.java
ch-dclarke: I think OAuth token might be better placed in the ouath package.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
ch-dclarke: If these are for debugging while the code is being built, it might be good to mark this with a //TODO remove the logging when we clean up.

src/main/java/uk/gov/ch/developer/docs/controller/developer/OAuthToken.java
Zain-Abbas1: Done

PR #58, opened 2020-03-12
Get User Profile

Zain-Abbas1:
- Create OAuthToken & UserProfileResponse class to map responses from api to objects
- Use access token to get user profile
- Then set user profile which is used by user bar to display details

__JIRA Ticket Number__
FA-606

### Type of change

* [ ] Bug fix
* [x] New feature
* [x] Breaking change

### Pull request checklist

### DIFF
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
@@ -1,6 +1,9 @@
 package uk.gov.ch.developer.docs.controller.developer;
 
 import java.net.URI;
+import java.util.HashMap;
+import java.util.Map;
+import javax.servlet.http.HttpServletRequest;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.MediaType;
 import org.springframework.stereotype.Controller;
@@ -15,22 +18,30 @@ import reactor.core.publisher.Mono;
 import uk.gov.ch.oauth.IIdentityProvider;
 import uk.gov.ch.oauth.IOauth;
 import uk.gov.ch.oauth.OAuthToken;
+import uk.gov.ch.oauth.UserProfileResponse;
 import uk.gov.companieshouse.logging.Logger;
 import uk.gov.companieshouse.logging.LoggerFactory;
+import uk.gov.companieshouse.session.Session;
+import uk.gov.companieshouse.session.SessionKeys;
+import uk.gov.companieshouse.session.handler.SessionHandler;
 
 @Controller
 @RequestMapping("${callback.url}")
 public class UserCallbackController {
 
     private static final Logger LOGGER = LoggerFactory.getLogger("docs.developer.ch.gov.uk");
+
     @Autowired
-    IIdentityProvider identityProvider;
+    private IIdentityProvider identityProvider;
+
     @Autowired
     private IOauth oauth;
 
+    private final WebClient webClient = WebClient.create();
+
     @GetMapping
     public String getCallback(@RequestParam("state") String state,
-            @RequestParam("code") String code) {
+            @RequestParam("code") String code, final HttpServletRequest httpServletRequest) {
         LOGGER.trace("Code:" + code);
         LOGGER.trace("State:" + state);
 
@@ -41,20 +52,88 @@ public class UserCallbackController {
             // Nonce values
         }
 
-        WebClient webClient = WebClient.create();
+        final Session chSession = (Session) httpServletRequest
+                .getAttribute(SessionHandler.CHS_SESSION_REQUEST_ATT_KEY);
 
-        webClient.post().uri(URI.create(identityProvider.getTokenUrl()))
-                .contentType(MediaType.APPLICATION_FORM_URLENCODED)
-                .accept(MediaType.APPLICATION_JSON)
-                .body(BodyInserters.fromPublisher(Mono.just(getPostRequestBody(code)),
-                        String.class))
-                .retrieve().bodyToMono(OAuthToken.class)
-                .subscribe(s -> LOGGER.trace("Access Token: " + s.getAccessToken()));// TODO remove
-                                                                                     // this log
-                                                                                     // line
+        Map<String, Object> signInData = new HashMap<>();
+
+        Mono<OAuthToken> oauthTokenResponse = getOAuthToken(code, webClient);
+
+        LOGGER.debug("Getting User Profile....");
+
+        UserProfileResponse userProfileResponse = getUserProfile(signInData, webClient, oauthTokenResponse);
+
+        if (userProfileResponse != null) {
+
+            LOGGER.debug("Processing Response");
+            setUserProfile(signInData, userProfileResponse);
+            signInData.put(SessionKeys.SIGNED_IN.getKey(), Integer.valueOf(1));
+            chSession.getData().put(SessionKeys.SIGN_IN_INFO.getKey(), signInData);
+
+        }
+
+        LOGGER.debug("Redirecting...");
 
         return ("redirect:" + identityProvider.getRedirectUriPage());// TODO redirect back to page
                                                                      // where sign-in was initiated
+    }
+
+    private UserProfileResponse getUserProfile(Map<String, Object> signInData, WebClient webClient,
+            Mono<OAuthToken> oauthTokenResponse) {
+        LOGGER.debug("Getting User Profile....");
+
+        OAuthToken oauthToken = oauthTokenResponse.block();
+
+        return requestUserProfile(signInData, webClient, oauthToken);
+    }
+
+    private UserProfileResponse requestUserProfile(Map<String, Object> signInData, WebClient webClient,
+            OAuthToken tokenResponse) {
+        setAccessToken(signInData, tokenResponse);
+        return webClient.get().uri(URI.create(identityProvider.getProfileUrl()))
+                .headers(h -> h.setBearerAuth(tokenResponse.getToken()))
+                .accept(MediaType.APPLICATION_JSON).exchange()
+                .flatMap(response -> response.bodyToMono(UserProfileResponse.class))
+                .block();
+    }
+
+    private Mono<OAuthToken> getOAuthToken(String code, WebClient webClient) {
+        LOGGER.debug("Getting OAuth Token....");
+        return webClient.post().uri(URI.create(identityProvider.getTokenUrl()))
+                .contentType(MediaType.APPLICATION_FORM_URLENCODED)
+                .accept(MediaType.APPLICATION_JSON).body(BodyInserters
+                        .fromPublisher(Mono.just(getPostRequestBody(code)), String.class))
+                .retrieve()
+                .bodyToMono(OAuthToken.class);
+    }
+
+    private void setUserProfile(Map<String, Object> signInData, UserProfileResponse userProfileResponse) {
+        LOGGER.debug("Setting User Profile...");
+        Map<String, Object> userProfileData = new HashMap<>();
+
+        userProfileData.put(SessionKeys.EMAIL.getKey(), userProfileResponse.getEmail());
+        userProfileData.put(SessionKeys.USER_ID.getKey(), userProfileResponse.getId());
+        userProfileData.put(SessionKeys.LOCALE.getKey(), userProfileResponse.getLocale());
+        userProfileData.put(SessionKeys.SCOPE.getKey(), userProfileResponse.getScope());
+        userProfileData.put(SessionKeys.FORENAME.getKey(), userProfileResponse.getForename());
+        userProfileData.put(SessionKeys.SURNAME.getKey(), userProfileResponse.getSurname());
+        userProfileData.put(SessionKeys.PERMISSIONS.getKey(), userProfileResponse.getPermissions());
+
+        LOGGER.debugContext("User Profile Data: ", "setUserProfile() Method", userProfileData);
+
+        signInData.put(SessionKeys.USER_PROFILE.getKey(), userProfileData);
+    }
+
+    private void setAccessToken(Map<String, Object> signInData, OAuthToken token) {
+        LOGGER.debug("Setting access token...");
+        Map<String, Object> accessTokenData = new HashMap<>();
+
+        accessTokenData.put(SessionKeys.ACCESS_TOKEN.getKey(), token.getToken());
+        accessTokenData.put(SessionKeys.EXPIRES_IN.getKey(), token.getExpiresIn());
+        accessTokenData.put(SessionKeys.REFRESH_TOKEN.getKey(), token.getRefreshToken());
+        accessTokenData.put(SessionKeys.TOKEN_TYPE.getKey(), token.getTokenType());
+
+        signInData.put(SessionKeys.ACCESS_TOKEN.getKey(), accessTokenData);
     }
 
     private String getPostRequestBody(String code) {
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
@@ -1,20 +1,24 @@
 package uk.gov.ch.developer.docs.controller.developer;
 
 import java.net.URI;
+import java.time.Duration;
 import java.util.HashMap;
 import java.util.Map;
 import javax.servlet.http.HttpServletRequest;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.MediaType;
+import org.springframework.http.ReactiveHttpOutputMessage;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.reactive.function.BodyInserter;
 import org.springframework.web.reactive.function.BodyInserters;
 import org.springframework.web.reactive.function.client.WebClient;
 import com.nimbusds.jose.Payload;
 import net.minidev.json.JSONObject;
 import reactor.core.publisher.Mono;
+import uk.gov.ch.developer.docs.DocsWebApplication;
 import uk.gov.ch.oauth.IIdentityProvider;
 import uk.gov.ch.oauth.IOauth;
 import uk.gov.ch.oauth.OAuthToken;
@@ -29,7 +33,7 @@ import uk.gov.companieshouse.session.handler.SessionHandler;
 @RequestMapping("${callback.url}")
 public class UserCallbackController {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger("docs.developer.ch.gov.uk");
+    private static final Logger LOGGER = LoggerFactory.getLogger(DocsWebApplication.APPLICATION_NAME_SPACE);
 
     @Autowired
     private IIdentityProvider identityProvider;
@@ -38,6 +42,7 @@ public class UserCallbackController {
     private IOauth oauth;
 
     private final WebClient webClient = WebClient.create();
+    private final Duration duration = Duration.ofSeconds(10L);
 
     @GetMapping
     public String getCallback(@RequestParam("state") String state,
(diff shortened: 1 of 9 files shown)


Zain-Abbas1:
There are some code smells/issues, however PR has been raised early in order for @ch-dclarke to view this before he leaves tomorrow.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -41,22 +52,90 @@ public String getCallback(@RequestParam("state") String state,
...
+
+        LOGGER.debug("Getting User Profile....");
+
+        UserProfileResponse userProfileResponse = getUserProfile(signInData, webClient, oauthTokenResponse);
+
+        if (userProfileResponse != null) {
+
+            LOGGER.debug("Processing Response");
+            setUserProfile(signInData, userProfileResponse);
+            signInData.put(SessionKeys.SIGNED_IN.getKey(), Integer.valueOf(1));
```
ch-dclarke: Please remove explicit boxing, It should be unnecessary.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -41,22 +52,90 @@ public String getCallback(@RequestParam("state") String state,
...
-                .accept(MediaType.APPLICATION_JSON)
-                .body(BodyInserters.fromPublisher(Mono.just(getPostRequestBody(code)),
-                        String.class))
-                .retrieve().bodyToMono(OAuthToken.class)
-                .subscribe(s -> LOGGER.trace("Access Token: " + s.getAccessToken()));// TODO remove
-                                                                                     // this log
-                                                                                     // line
+        Map<String, Object> signInData = new HashMap<>();
+
+        Mono<OAuthToken> oauthTokenResponse = getOAuthToken(code, webClient);
```
ch-dclarke: For this, use `.block()` with a timeout inside `getOAuthToken(...)` So that you have a POJO `OAuthToken` rather than a delayed instance.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -15,22 +18,30 @@
...
     private static final Logger LOGGER = LoggerFactory.getLogger("docs.developer.ch.gov.uk");
+
     @Autowired
-    IIdentityProvider identityProvider;
+    private IIdentityProvider identityProvider;
+
     @Autowired
     private IOauth oauth;
 
+    private final WebClient webClient = WebClient.create();
```
ch-dclarke: I think `webClient` should not be a member variable, it should be a local variable within its usages.
Each time you `create()` any previous creations will need to be disposed of, or the code will leak connection handles.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
ch-dclarke: Move `webclient` inside the `getOauthToken()`, create it and destroy it there.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -41,22 +52,90 @@ public String getCallback(@RequestParam("state") String state,
...
 
         return ("redirect:" + identityProvider.getRedirectUriPage());// TODO redirect back to page
                                                                      // where sign-in was initiated
     }
 
+    private UserProfileResponse getUserProfile(Map<String, Object> signInData, WebClient webClient,
+            Mono<OAuthToken> oauthTokenResponse) {
+        LOGGER.debug("Getting User Profile....");
+
+        OAuthToken oauthToken = oauthTokenResponse.block();
```
ch-dclarke: `.block` needs a timeout.
It would be better to move the `oauthTokenResponse.block()` into `getOauthToken(...)` so that you have a POJO token coming in.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -41,22 +52,90 @@ public String getCallback(@RequestParam("state") String state,
...
+
+        return requestUserProfile(signInData, webClient, oauthToken);
+    }
+
+    private UserProfileResponse requestUserProfile(Map<String, Object> signInData, WebClient webClient,
```
ch-dclarke: Use a local WebClient and dispose of it, rather than using one that is passed in.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -41,22 +52,90 @@ public String getCallback(@RequestParam("state") String state,
...
+        Map<String, Object> userProfileData = new HashMap<>();
+
+        userProfileData.put(SessionKeys.EMAIL.getKey(), userProfileResponse.getEmail());
+        userProfileData.put(SessionKeys.USER_ID.getKey(), userProfileResponse.getId());
+        userProfileData.put(SessionKeys.LOCALE.getKey(), userProfileResponse.getLocale());
+        userProfileData.put(SessionKeys.SCOPE.getKey(), userProfileResponse.getScope());
+        userProfileData.put(SessionKeys.FORENAME.getKey(), userProfileResponse.getForename());
+        userProfileData.put(SessionKeys.SURNAME.getKey(), userProfileResponse.getSurname());
+        userProfileData.put(SessionKeys.PERMISSIONS.getKey(), userProfileResponse.getPermissions());
+
```
ch-dclarke: Why not unwrap the UserProfile response to a single UserProfile entity and put that in, with a single key. Then new properties aren't missed, and `UserProfile` won't be fragmented. Consumers will be able to take the UserProfile directly, without rebuilding it field by field.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -41,22 +52,90 @@ public String getCallback(@RequestParam("state") String state,
...
+        return requestUserProfile(signInData, webClient, oauthToken);
+    }
+
+    private UserProfileResponse requestUserProfile(Map<String, Object> signInData, WebClient webClient,
+            OAuthToken tokenResponse) {
+        setAccessToken(signInData, tokenResponse);
+        return webClient.get().uri(URI.create(identityProvider.getProfileUrl()))
+                .headers(h -> h.setBearerAuth(tokenResponse.getToken()))
+                .accept(MediaType.APPLICATION_JSON).exchange()
```
CH-JonathanBlake: Please split these consistently to one method per line. (On the web client at least)

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -41,22 +52,90 @@ public String getCallback(@RequestParam("state") String state,
...
+        LOGGER.debug("Getting User Profile....");
+
+        OAuthToken oauthToken = oauthTokenResponse.block();
+
+        return requestUserProfile(signInData, webClient, oauthToken);
+    }
+
+    private UserProfileResponse requestUserProfile(Map<String, Object> signInData, WebClient webClient,
+            OAuthToken tokenResponse) {
+        setAccessToken(signInData, tokenResponse);
```
CH-JonathanBlake: Why are we doing this in this method? I think it would be better that if signInData is not needed in this method I'm worried this might obscure the trace of building up our user data.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -41,22 +52,90 @@ public String getCallback(@RequestParam("state") String state,
...
+        Map<String, Object> accessTokenData = new HashMap<>();
+
+        accessTokenData.put(SessionKeys.ACCESS_TOKEN.getKey(), token.getToken());
+        accessTokenData.put(SessionKeys.EXPIRES_IN.getKey(), token.getExpiresIn());
+        accessTokenData.put(SessionKeys.REFRESH_TOKEN.getKey(), token.getRefreshToken());
+        accessTokenData.put(SessionKeys.TOKEN_TYPE.getKey(), token.getTokenType());
+
+        signInData.put(SessionKeys.ACCESS_TOKEN.getKey(), accessTokenData);
+    }
+
```
ch-dclarke: Most of the `getPostRequestBody` looks like it should be delegated to the identity provider or OAuth instance.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -41,22 +52,90 @@ public String getCallback(@RequestParam("state") String state,
...
+        LOGGER.debug("Setting User Profile...");
+        Map<String, Object> userProfileData = new HashMap<>();
+
+        userProfileData.put(SessionKeys.EMAIL.getKey(), userProfileResponse.getEmail());
+        userProfileData.put(SessionKeys.USER_ID.getKey(), userProfileResponse.getId());
+        userProfileData.put(SessionKeys.LOCALE.getKey(), userProfileResponse.getLocale());
+        userProfileData.put(SessionKeys.SCOPE.getKey(), userProfileResponse.getScope());
+        userProfileData.put(SessionKeys.FORENAME.getKey(), userProfileResponse.getForename());
+        userProfileData.put(SessionKeys.SURNAME.getKey(), userProfileResponse.getSurname());
+        userProfileData.put(SessionKeys.PERMISSIONS.getKey(), userProfileResponse.getPermissions());
```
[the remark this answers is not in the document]
CH-JonathanBlake: And the same thing for access token

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
Zain-Abbas1: How long would you suggest the timeout be?

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
Zain-Abbas1: After yesterday's s conversation we agreed this should remain here, subscriptions are disposed of not the actual web client.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
Zain-Abbas1: As discussed earlier on - this was done in order to comply with the java-session-handler and how it unwraps and builds the user profile. Therefore it was necessary to add each field to the map here, which is then read field by field and set by the java-session-handler.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -41,22 +52,90 @@ public String getCallback(@RequestParam("state") String state,
...
+        UserProfileResponse userProfileResponse = getUserProfile(signInData, webClient, oauthTokenResponse);
+
+        if (userProfileResponse != null) {
+
+            LOGGER.debug("Processing Response");
+            setUserProfile(signInData, userProfileResponse);
+            signInData.put(SessionKeys.SIGNED_IN.getKey(), Integer.valueOf(1));
+            chSession.getData().put(SessionKeys.SIGN_IN_INFO.getKey(), signInData);
+
+        }
```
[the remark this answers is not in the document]
Zain-Abbas1: We have a generic error handling ticket, this will have to be a part of it.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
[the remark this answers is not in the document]
Zain-Abbas1: I've done this, but I am passing in the signInData as well - not sure if I should do that while I'm setting the user profile or at the controller level.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -41,30 +57,63 @@ public String getCallback(@RequestParam("state") String state,
...
+    }
 
-        return sb.toString();
+    private OAuthToken getOAuthToken(String code) {
+        LOGGER.debug("Getting OAuth Token");
+        
+        final URI tokenUrl = URI.create(identityProvider.getTokenUrl());
+        final Mono<String> postRequest = Mono.just(identityProvider.getPostRequestBody(code));
+        final BodyInserter<Mono<String>, ReactiveHttpOutputMessage> bodyInserter = BodyInserters
+        .fromPublisher(postRequest, String.class);
```
CH-JonathanBlake: Indent this? Did you run the code reformatter?

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -41,30 +57,63 @@ public String getCallback(@RequestParam("state") String state,
...
-                .append("&client_secret=").append(identityProvider.getClientSecret())
-                .append("&redirect_uri=").append(identityProvider.getRedirectUri())
-                .append("&grant_type=authorization_code");
+    private UserProfileResponse getUserProfile(OAuthToken tokenResponse) {
+        LOGGER.debug("Requesting User Profile");
+        
+        final URI profileUrl = URI.create(identityProvider.getProfileUrl());
+        
+        return webClient.get()
+                .uri(profileUrl)
```
ch-dclarke: It might be better to abstract this business logic, into the OAuth2 class, as it really should know what is going on. It also means that if the logic is changed in the OAuth processes, we don't need to change controllers.
Compare with the SignIn controller for preparation and retrieving of data.

src/main/java/uk/gov/ch/oauth/UserProfileResponse.java
```diff
@@ -0,0 +1,70 @@
...
+        userProfileData.put(SessionKeys.EMAIL.getKey(), getEmail());
+        userProfileData.put(SessionKeys.USER_ID.getKey(), getId());
+        userProfileData.put(SessionKeys.LOCALE.getKey(), getLocale());
+        userProfileData.put(SessionKeys.SCOPE.getKey(), getScope());
+        userProfileData.put(SessionKeys.FORENAME.getKey(), getForename());
+        userProfileData.put(SessionKeys.SURNAME.getKey(), getSurname());
+        userProfileData.put(SessionKeys.PERMISSIONS.getKey(), getPermissions());
+
+        signInData.put(SessionKeys.USER_PROFILE.getKey(), userProfileData);
+    }
```
CH-JonathanBlake: This looks like a very strange method for an object oriented design.
I don't think this should be called a set method, and I think that it should actually return the userProfileData map and not accept an input.
And the signInData.put would wrap this method:
`signInData.put(SessionKeys.USER_PROFILE.getKey(), userProfileData.convertToMap());`

src/main/java/uk/gov/ch/oauth/OAuthToken.java
```diff
@@ -1,50 +1,49 @@
...
+
+        accessTokenData.put(SessionKeys.ACCESS_TOKEN.getKey(), getToken());
+        accessTokenData.put(SessionKeys.EXPIRES_IN.getKey(), getExpiresIn());
+        accessTokenData.put(SessionKeys.REFRESH_TOKEN.getKey(), getRefreshToken());
+        accessTokenData.put(SessionKeys.TOKEN_TYPE.getKey(), getTokenType());
 
-    public void setAccessToken(String accessToken) {
-        this.accessToken = accessToken;
+        signInData.put(SessionKeys.ACCESS_TOKEN.getKey(), accessTokenData);
     }
```
CH-JonathanBlake: Same comment as made on the UserProfileResponse.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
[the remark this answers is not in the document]
ch-dclarke: Probably, both belong together in the OAuth classes.

src/main/java/uk/gov/ch/oauth/IdentityProvider.java
```diff
@@ -95,6 +95,16 @@ public String getAuthorisationUrl(final String state) {
         return url;
     }
 
+    @Override
+    public String getPostRequestBody(String code) {
+        final StringBuilder sb = new StringBuilder();
+        sb.append("code=").append(code).append("&client_id=").append(getClientId())
+                .append("&client_secret=").append(getClientSecret()).append("&redirect_uri=")
+                .append(getRedirectUri()).append("&grant_type=authorization_code");
```
CH-JonathanBlake: Should authorisation_code be a constant?
Would this be more readable as
`return String.format("code={}&client_id={}&client_secret={}&redirect_uri={}&grant_type=authorization_code",
     code,
     getClientId(),
    getClientSecret(),
    getRedirectUri();
`
Except maybe with authorization code also being a constant

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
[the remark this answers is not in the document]
CH-JonathanBlake: Oops, effectively tripled this discussion.
I think 
'signInData.put(SessionKeys.USER_PROFILE.getKey(), userProfileData.convertToMap());'
would look better. 
From an object oriented point of view, the UserProfileResponse class should be doing things that affect it, not arbitrary objects.

src/main/java/uk/gov/ch/oauth/OAuthToken.java
CH-JonathanBlake: This is very not a Setter.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -1,36 +1,38 @@
...
-    IIdentityProvider identityProvider;
+    private IIdentityProvider identityProvider;
+
     @Autowired
     private IOauth oauth;
 
     @GetMapping
     public String getCallback(@RequestParam("state") String state,
-            @RequestParam("code") String code) {
+            @RequestParam("code") String code, final HttpServletRequest httpServletRequest) {
```
ch-dclarke: Should the serverlet request have an annontation?

Zain-Abbas1: What annotation?

ch-dclarke: Good question... as it is already auto wired, I'm not sure.

src/main/java/uk/gov/ch/oauth/OAuthToken.java
Zain-Abbas1: Renamed

src/main/java/uk/gov/ch/oauth/IIdentityProvider.java
```diff
@@ -49,4 +51,6 @@
     String getRedirectUriPage();
     
     String getAccountLocalUrl();
+    
+    String getGrantType();
```
ch-dclarke: Does grant type get used anywhere outside of the identity provider? If not, drop it from the interface.

src/main/java/uk/gov/ch/oauth/IdentityProvider.java
```diff
@@ -108,5 +120,9 @@ public String getRedirectUriPage() {
     public String getAccountLocalUrl() {
         return accountLocalUrl;
     }
-
+    
+    @Override
```
ch-dclarke: No reason to be public.

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -101,4 +121,57 @@ private String getSessionNonce() {
...
+        LOGGER.debug("Requesting User Profile");
+
+        final OAuthToken oauthToken = getOAuthToken(code);
+        final WebClient webClient = WebClient.create();
+        final URI profileUrl = URI.create(identityProvider.getProfileUrl());
+
+        final UserProfileResponse userProfile = getUserProfileResponse(oauthToken, webClient,
+                profileUrl);
+
+        if (userProfile != null) {
```
ch-dclarke: Can never be null, as a result of the `.block()` call to reach it.

src/main/java/uk/gov/ch/oauth/IIdentityProvider.java
Zain-Abbas1: Much like the other fields they are only used within the Identity provider.

src/main/java/uk/gov/ch/oauth/IdentityProvider.java
Zain-Abbas1: Follows suit of other fields

src/main/java/uk/gov/ch/oauth/IIdentityProvider.java
Zain-Abbas1: For future proofing, we might handle different grant types which is why I put it in the interface.

ch-dclarke: It would still only be used internally, so please drop it out, for now.

Zain-Abbas1: The same could be said for some of the other fields, I don't think it makes a difference.

PR #59, opened 2020-03-18
feature/fa-417 Updated user and nav bar

CH-JonathanBlake:
Updated user and nav bar to latest prototype version.
__JIRA Ticket Number__
https://companieshouse.atlassian.net/browse/FA-417
### Type of change

* [X] Bug fix
* [ ] New feature
* [ ] Breaking change

### Pull request checklist

### DIFF
--- a/src/main/java/uk/gov/ch/developer/docs/config/NavBarConfig.java
+++ b/src/main/java/uk/gov/ch/developer/docs/config/NavBarConfig.java
@@ -44,14 +44,6 @@ public class NavBarConfig {
         documentation.add("How to add an API key", authenticationUrl);
         documentation.add("Developer Guidelines", devGuideURL);
 
-        NavItemList manageAccounts = model
-                .addHeading("Manage account", DisplayRestrictions.USER_REQUIRED);
-        manageAccounts.add("Manage Profile", homeURL);
-        manageAccounts.add("Change Password", homeURL);
-
-        NavItemList help = model.addHeading("Help", DisplayRestrictions.USER_REQUIRED);
-        help.add("Developer Hub Forum", homeURL);
-
         return model;
     }
 }
--- a/src/main/resources/templates/fragments/userBar.html
+++ b/src/main/resources/templates/fragments/userBar.html
@@ -9,15 +9,9 @@
   <nav class="content">
     <ul class="mobile-hidden" id="navigation">
       <li>
-        <ul class="mobile-hidden"
-            th:if="${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}">
+        <ul th:if="${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}"
+            th:remove="tag">
           <li class="user" id="signed-in-user" th:text="${userModel.getId() + ':'}"></li>
-          <li><a
-              class="piwik-event"
-              data-event-id="Your details - top menu" id="your-details"
-              th:href="@{{chsAccountUrl}/user/account(chsAccountUrl=${@environment.getProperty('account.url')})}">Your
-            details
-          </a></li>
           <li><a
               class="piwik-event"
               data-event-id="Your Applications" id="your-applications"
@@ -29,12 +23,13 @@
           </a></li>
         </ul>
       </li>
-      <li th:if="!${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}">
-        <!--  -->
+      <li th:if="!${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}"
+          th:remove="tag">
         <a class="piwik-event govuk-link" data-event-id="Sign In" href="/signin" id="user-signin">Sign
           in /
           Register</a>
       </li>
+      <li><a class="govuk-link" href="http://www.example.com">Developer hub forum</a></li>
     </ul>
   </nav>
 </div>
--- a/src/main/resources/templates/fragments/userBar.html
+++ b/src/main/resources/templates/fragments/userBar.html
@@ -7,28 +7,24 @@
 <body>
 <div class="js-toggle-nav" id="global-nav" th:fragment="userBar">
   <nav class="content">
-    <ul class="mobile-hidden" id="navigation">
-      <li>
-        <ul th:if="${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}"
-            th:remove="tag">
+    <div
+        th:with="loggedIn=${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}">
+      <ul class="mobile-hidden" id="navigation" th:switch="${loggedIn}">
+        <ul th:case="true" th:remove="tag">
           <li class="user" id="signed-in-user" th:text="${userModel.getId()}"></li>
-          <li><a class="govuk-link"
-                 th:href="@{{forum}(forum=${@environment.getProperty('developerforum.url')})}">Developer
-            hub forum</a></li>
           <li><a class="piwik-event" data-event-id="Sign out" href="/signout" id="user-signout">Sign
             out
           </a></li>
         </ul>
-      </li>
-      <ul th:if="!${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}"
-          th:remove="tag">
+        <ul th:case="false" th:remove="tag">
+          <li><a class="piwik-event govuk-link" data-event-id="Sign In" href="/signin"
+                 id="user-signin">Sign in / Register</a></li>
+        </ul>
         <li><a class="govuk-link"
                th:href="@{{forum}(forum=${@environment.getProperty('developerforum.url')})}">Developer
           hub forum</a></li>
-        <li><a class="piwik-event govuk-link" data-event-id="Sign In" href="/signin"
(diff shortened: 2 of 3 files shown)


src/main/resources/templates/fragments/userBar.html
```diff
@@ -29,12 +23,13 @@
...
       </li>
-      <li th:if="!${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}">
-        <!--  -->
+      <li th:if="!${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}"
+          th:remove="tag">
         <a class="piwik-event govuk-link" data-event-id="Sign In" href="/signin" id="user-signin">Sign
           in /
           Register</a>
       </li>
+      <li><a class="govuk-link" href="http://www.example.com">Developer hub forum</a></li>
```
Zain-Abbas1: Please use the developer hub forum link

src/main/resources/templates/fragments/userBar.html
```diff
@@ -9,15 +9,9 @@
...
+        <ul th:if="${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}"
+            th:remove="tag">
           <li class="user" id="signed-in-user" th:text="${userModel.getId() + ':'}"></li>
-          <li><a
-              class="piwik-event"
-              data-event-id="Your details - top menu" id="your-details"
-              th:href="@{{chsAccountUrl}/user/account(chsAccountUrl=${@environment.getProperty('account.url')})}">Your
-            details
-          </a></li>
           <li><a
```
Zain-Abbas1: Please remove this link, when user is signed in the only information that is displayed is - User ID, Dev Forum, Sign out - in that order

src/main/resources/templates/fragments/userBar.html
ch-dclarke: We don't have one yet, so the example.com place holders are there to make them easy to find with valid links.

src/main/java/uk/gov/ch/developer/docs/config/NavBarConfig.java
```diff
@@ -44,14 +44,6 @@ public NavBarModelBuilder getNavBarModelBuilder() {
         documentation.add("How to add an API key", authenticationUrl);
         documentation.add("Developer Guidelines", devGuideURL);
 
-        NavItemList manageAccounts = model
-                .addHeading("Manage account", DisplayRestrictions.USER_REQUIRED);
-        manageAccounts.add("Manage Profile", homeURL);
-        manageAccounts.add("Change Password", homeURL);
-
-        NavItemList help = model.addHeading("Help", DisplayRestrictions.USER_REQUIRED);
```
ch-dclarke: It seems a little of that the help is only restricted to logged in users.

src/main/resources/templates/fragments/userBar.html
Zain-Abbas1: Are we not linking it to the current dev hub forum that is live?

ch-dclarke: In general we have not linked to old ones that we should supersede. It will make it harder to spot errors in our coding.
If the link was resourced, then we should be safer, when it's injected into the ThymeLeaf view.

Zain-Abbas1: As per my understanding we are not changing the developer hub forum, it is staying the same. So we should link to the live forum.

src/main/resources/templates/fragments/userBar.html
CH-JonathanBlake: Okay sorry I didn't realise applications also had to go.

src/main/java/uk/gov/ch/developer/docs/config/NavBarConfig.java
CH-JonathanBlake: This was removed.

ch-dclarke: My mistake, sorry.

src/main/resources/templates/fragments/userBar.html
ch-dclarke: The Forum is on our tickets, just much lower priority than anything else, but setting it in the properties would make the change easy, if the dev-hub is removed from scope.

CH-JonathanBlake: Micheal was able to confirm that the current live address should be used.

CH-JonathanBlake:
Changes made, requesting rereview

src/main/resources/templates/fragments/userBar.html
```diff
@@ -9,32 +9,21 @@
...
       </li>
-      <li th:if="!${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}">
-        <!--  -->
+      <li th:if="!${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}"
+          th:remove="tag">
         <a class="piwik-event govuk-link" data-event-id="Sign In" href="/signin" id="user-signin">Sign
           in /
           Register</a>
       </li>
+      <li><a class="govuk-link" href="https://forum.aws.chdev.org/ ">Developer hub forum</a></li>
```
ch-dclarke: Apart from linking to the old dev-forum, I think the extra space after the url may confuse some technologies.

CH-JonathanBlake: Worked fine when I tried, but I'll fix incase its an IE thing.
Old dev forum is correct, I verified with Mike.

src/main/resources/templates/fragments/userBar.html
```diff
@@ -9,32 +9,21 @@
...
       </li>
-      <li th:if="!${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}">
-        <!--  -->
+      <li th:if="!${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}"
+          th:remove="tag">
         <a class="piwik-event govuk-link" data-event-id="Sign In" href="/signin" id="user-signin">Sign
           in /
           Register</a>
       </li>
+      <li><a class="govuk-link" href="https://forum.aws.chdev.org/">Developer hub forum</a></li>
```
Zain-Abbas1: Please see above comment - Michael Gearon suggested we should have the dev hub forum link before Sign Out link. This mirrors current live developer hub.

CH-JonathanBlake: Changed.

src/main/resources/templates/fragments/userBar.html
```diff
@@ -9,32 +9,25 @@
...
       </li>
-      <li th:if="!${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}">
-        <!--  -->
-        <a class="piwik-event govuk-link" data-event-id="Sign In" href="/signin" id="user-signin">Sign
-          in /
-          Register</a>
-      </li>
+      <ul th:if="!${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}"
+          th:remove="tag">
+        <li><a class="govuk-link"
```
Zain-Abbas1: Sorry to be a pain Jon, but it should be Sign in/Register first then dev hub forum. Please check prototype.

src/main/resources/templates/fragments/userBar.html
```diff
@@ -9,32 +9,25 @@
...
           </a></li>
         </ul>
       </li>
-      <li th:if="!${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}">
-        <!--  -->
-        <a class="piwik-event govuk-link" data-event-id="Sign In" href="/signin" id="user-signin">Sign
-          in /
-          Register</a>
-      </li>
+      <ul th:if="!${T(uk.gov.ch.developer.docs.models.user.UserModel).isUserSignedIn(userModel)}"
```
Zain-Abbas1: Do we need this if statement, shouldn't this be the default unless the first if-statement is fulfilled? Can we look at using if-else, or switch statements if removing this doesn't work for whatever reasonn.

CH-JonathanBlake: this is thymeleaf not java... It doesn't have if else.
they have this odd choose thing? I'll see what I can do but might just ignore you on this.

Zain-Abbas1: I realise its thymeleaf - there are alternatives I listed above such as using a switch statement  or if-unless (although i'm not to keen on that). Don't see the point in evaluating the statement twice if you're not signed in.

CH-JonathanBlake: Even with a choose (or case statement (as switches are called in thymeleaf. Because its not java.)) we'd still be evaluating it multiple times because its not an if else.
It has to evaluate with the possibility that it could be either, both or neither...

CH-JonathanBlake: https://www.thymeleaf.org/whatsnew20.html#swit
I was mistaken. Switch statements exit as soon as they find a match.

src/main/resources/templates/fragments/userBar.html
CH-JonathanBlake: After speaking with michael gearon, he has agreed that we should use a layout that is consistent between logged in and logged out, that places the developer forum on the far right.

Zain-Abbas1:
Looks good to me 👍

PR #61, opened 2020-03-30
Implement Signout

Zain-Abbas1:
Add signout controller that redirects to chs /signout

__JIRA Ticket Number__
FA-390

### Type of change

* [ ] Bug fix
* [x] New feature
* [ ] Breaking change

### Pull request checklist

### DIFF
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
@@ -1,0 +1,56 @@
+package uk.gov.ch.developer.docs.controller.developer;
+
+import java.io.IOException;
+import java.util.Map;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Controller;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    @GetMapping
+    public void doSignOut(HttpServletRequest httpServletRequest,
+            HttpServletResponse httpServletResponse) throws IOException {
+
+        LOGGER.debug("Processing redirect to sign out");
+
+        httpServletResponse.sendRedirect(signOutUrl);
+    }
```
ch-dclarke: Since this controller only has one method, it might be better to give the `@GetMapping` the path, rather than the whole controller. This would reduce the complexity a little.

Also parameters that should not be changes, should be marked as final.

ch-dclarke:
Can you include some unit tests?

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
Zain-Abbas1: Done

PR #67, opened 2020-04-16
Feature/fa 707 javadoc auth classes

ch-dclarke:
Added JavaDoc to classes extended from ch models
Updated POM to fix the version of compilations, in case of later compiler installs

__JIRA Ticket Number__
FA-707

### Type of change

* [x] Bug fix
* [ ] New feature
* [ ] Breaking change

### Pull request checklist


CH-JonathanBlake:
Why are we pushing ideas?

src/main/java/uk/gov/ch/oauth/UserProfileResponse.java
```diff
@@ -48,13 +60,22 @@ public void setSurname(String surname) {
 
     @JsonProperty("permissions")
     @Override
-    public void setPermissions(Map<String, Boolean> permissions) {
+    public void setPermissions(final Map<String, Boolean> permissions) {
```
CH-JonathanBlake, line 63: Strange that you've only done this on this setter. Could you explain? When these methods are just wrappers should we mimic the superclasses finalisations?

ch-dclarke: We don't need to provide annotations to the getters as we don't send the values over the wire.

PR #68, opened 2020-04-16
Introduced code where nonce value is updated to an empty value

Ngriffiths1:
This was done to match the perl code that we are refactoring into Java

Tests for this class will be picked up in a later ticket FA-687

__JIRA Ticket Number__
FA-704

### Type of change

* [ ] Bug fix
*  New feature
* [ ] Breaking change

### Pull request checklist

*  The code follows our [coding standards](https://github.com/companieshouse/styleguides/blob/master/java.md)

### DIFF
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
@@ -1,13 +1,13 @@
 package uk.gov.ch.developer.docs.controller.developer;
 
+import com.nimbusds.jose.Payload;
 import javax.servlet.http.HttpServletRequest;
+import net.minidev.json.JSONObject;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
-import com.nimbusds.jose.Payload;
-import net.minidev.json.JSONObject;
 import uk.gov.ch.developer.docs.DocsWebApplication;
 import uk.gov.ch.oauth.IIdentityProvider;
 import uk.gov.ch.oauth.IOauth;
@@ -39,7 +39,8 @@ public class UserCallbackController {
                 .getAttribute(SessionHandler.CHS_SESSION_REQUEST_ATT_KEY);
 
         final String returnedNonce = getNonceFromState(state);
-        if (!oauth.oauth2VerifyNonce(returnedNonce, chSession)) {
+        final String sessionNonce = oauth.getSessionNonce(chSession);
+        if (!oauth.oauth2VerifyNonce(returnedNonce, sessionNonce)) {
             LOGGER.error("Invalid nonce value in state during oauth2 callback");
             // return "redirect:/"; TODO redirect will not work, needs to be addressed for unmatched
             // Nonce values
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
@@ -35,16 +35,15 @@ public class UserCallbackController {
             @RequestParam("code") String code, final HttpServletRequest httpServletRequest) {
         LOGGER.trace("Code:" + code);
         LOGGER.trace("State:" + state);
+        final Session chSession = (Session) httpServletRequest
+                .getAttribute(SessionHandler.CHS_SESSION_REQUEST_ATT_KEY);
 
         final String returnedNonce = getNonceFromState(state);
-        if (!oauth.oauth2VerifyNonce(returnedNonce)) {
+        if (!oauth.oauth2VerifyNonce(returnedNonce, chSession)) {
             LOGGER.error("Invalid nonce value in state during oauth2 callback");
             // return "redirect:/"; TODO redirect will not work, needs to be addressed for unmatched
             // Nonce values
         }
-
-        final Session chSession = (Session) httpServletRequest
-                .getAttribute(SessionHandler.CHS_SESSION_REQUEST_ATT_KEY);
 
         LOGGER.debug("Getting User Profile");
 
--- a/src/main/java/uk/gov/ch/oauth/Oauth2.java
+++ b/src/main/java/uk/gov/ch/oauth/Oauth2.java
@@ -114,7 +114,7 @@ public class Oauth2 implements IOauth {
 
     /**
      * Extract the OAuth2 Nonce from the current session
-     * Removes Nonce value so can be validated fresh and first obtaining the value
+     * Removes Nonce value so can be validated once and replay attacks are more difficult
      *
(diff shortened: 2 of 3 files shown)


src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -35,17 +35,16 @@ public String getCallback(@RequestParam("state") String state,
             @RequestParam("code") String code, final HttpServletRequest httpServletRequest) {
         LOGGER.trace("Code:" + code);
         LOGGER.trace("State:" + state);
+        final Session chSession = (Session) httpServletRequest
+                .getAttribute(SessionHandler.CHS_SESSION_REQUEST_ATT_KEY);
 
         final String returnedNonce = getNonceFromState(state);
-        if (!oauth.oauth2VerifyNonce(returnedNonce)) {
+        if (!oauth.oauth2VerifyNonce(returnedNonce, chSession)) {
```
ch-dclarke, line 42: Rather than passing the session through, it would be better to call `getSessionNonce(chSession)` outside then pass the result into `oauth2VerifyNonce(...)`

On the lines of:

```
final String returnedNonce = getNonceFromState(state);
final String sessionNonce = oauth2.getAndDestroySessionNonce(chSession);
if (!oauth.oauth2VerifyNonce(returnedNonce, sessionNonce)) {

```

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -111,11 +111,13 @@ public boolean oauth2VerifyNonce(final String nonce) {
...
      */
-    private String getSessionNonce() {
+    private String getSessionNonce(Session chSession) {
         String oauth2Nonce = null;
         try {
-            final Map<String, Object> data = sessionService.getSessionDataFromContext();
+            final Map<String, Object> data = chSession.getData();
             oauth2Nonce = (String) data.getOrDefault(SessionKeys.NONCE.getKey(), null);
+            data.put(SessionKeys.NONCE.getKey(), "");
+            LOGGER.debug("Verify nonce against expected nonce from session");
```
ch-dclarke, line 120: I don't understand this log message in this context.

ch-dclarke, line 120: For clarity, it might be better to have a `destroyNonce()` method that you call here. That gives a better separation of concerns.

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -98,10 +98,10 @@ public Payload oauth2DecodeState(final String state) {
         return payload;
     }
 
-    public boolean oauth2VerifyNonce(final String nonce) {
+    public boolean oauth2VerifyNonce(final String nonce, Session chSession) {
```
ch-dclarke, line 101: Second Parameter should also be final.

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -111,11 +111,13 @@ public boolean oauth2VerifyNonce(final String nonce) {
      *
      * @return The Nonce String from within the session or null if not found.
      */
-    private String getSessionNonce() {
+    private String getSessionNonce(Session chSession) {
```
ch-dclarke, line 114: Make the session parameter final.

src/main/java/uk/gov/ch/oauth/Oauth2.java
Ngriffiths1: The debug message was taken from the Perl code example in https://github.com/companieshouse/CH-MojoX-SignIn/blob/develop/lib/CH/MojoX/SignIn/Plugin/OAuth2.pm#L46. If you have any recommendations for a better debug message that would be great.

ch-dclarke: I'd suggest that it isn't verifying in this method, so if you do need a debug message there, then maybe "Obtained nonce from session".

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
CH-JonathanBlake: I disagree. I'm working on a ticket to remove generating nonces from the signin controller after all.

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -111,11 +111,12 @@ public boolean oauth2VerifyNonce(final String nonce) {
      *
      * @return The Nonce String from within the session or null if not found.
      */
```
Zain-Abbas1, line 113: JavaDoc needs updating

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -98,10 +98,10 @@ public Payload oauth2DecodeState(final String state) {
         return payload;
     }
```
Zain-Abbas1, line 100: JavaDoc needs adding

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -111,11 +111,12 @@ public boolean oauth2VerifyNonce(final String nonce) {
...
      * @return The Nonce String from within the session or null if not found.
      */
-    private String getSessionNonce() {
+    public String getSessionNonce(final Session chSession) {
         String oauth2Nonce = null;
         try {
-            final Map<String, Object> data = sessionService.getSessionDataFromContext();
-            oauth2Nonce = (String) data.getOrDefault(SessionKeys.NONCE.getKey(), null);
+            final Map<String, Object> data = chSession.getData();
+            oauth2Nonce = (String) data.remove(SessionKeys.NONCE.getKey());
```
Zain-Abbas1, line 118: If we're removing the nonce, then we must make sure we're accounting for the possibility of a potential null pointer upstream. I would feel more comfortable updating the value to a blank which is a direct copy of the perl implementation, however as long as we have thought about a potential null pointer then I suppose it's fine.

Ngriffiths1:
> What has been decided about the unit tests for this?

I believe the unit tests for this class will be picked up in FA-687

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -98,24 +98,32 @@ public Payload oauth2DecodeState(final String state) {
...
         boolean retval = false;
         if (nonce != null) {
-            retval = nonce.equals(getSessionNonce());
+            retval = nonce.equals(sessionNonce);
         }
         return retval;
     }
 
     /**
      * Extract the OAuth2 Nonce from the current session
```
Zain-Abbas1, line 116: I would delete this now as method function has changed.

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -98,24 +98,32 @@ public Payload oauth2DecodeState(final String state) {
...
+     * Removes Nonce value so can be validated fresh and first obtaining the value
```
Zain-Abbas1, line 117: Update this to read `Removes Nonce so value can only be verified once`

PR #70, opened 2020-04-21
Feature/fa 708 generate nonce oauth

CH-JonathanBlake:

Removed Nonces and dependence on docs beans from the oauth code.
Added tests for Token classes.

JIRA Ticket Number

https://companieshouse.atlassian.net/browse/FA-708
### Type of change

* [?] Bug fix
* [ ] New feature
* [X] Breaking change

### Pull request checklist


src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -30,15 +34,15 @@
 public class Oauth2 implements IOauth {
 
     private static final Logger LOGGER = LoggerFactory.getLogger("docs.developer.ch.gov.uk");
-    final IIdentityProvider identityProvider;
+    private final IIdentityProvider identityProvider;
+    private final SessionUtils sessionUtils;
     private final Duration timeoutDuration = Duration.ofSeconds(10L);
+    private final NonceGenerator nonceGenerator = new NonceGenerator();
```
ch-dclarke, line 40: Is it worth considering auto wiring the `NonceGenerator` to give us the testing option of known Nonces very easily?

src/test/java/uk/gov/ch/oauth/nonce/NonceGeneratorTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Test
+    @DisplayName("Generate Nonce creates unique values, brute force test")
+    void generateNonceGeneratesUniqueValues() {
+        List<String> nonces = new ArrayList<>();
+        for (int i = 0; i < 100; i++) {
+            String newNonce = nonceGenerator.generateNonce();
+            System.out.println(newNonce);
+            assertFalse(nonces.contains(newNonce));
+            nonces.add(newNonce);
```
ch-dclarke, line 74: Please remove the `System.out` as unnecessary.

The performance of checking each value as you go along in a `List` is very inefficient.

I would suggest an ordered `Set` such as a `TreeSet` where the search time each time is O(logN) as opposed to O(N) and the `boolean add(...)` method also gives a test result for whether it has encountered the value before.

src/test/java/uk/gov/ch/oauth/nonce/NonceGeneratorTest.java
```diff
@@ -0,0 +1,77 @@
...
+        // prevent code coverage marking this as done.
+        String retNonce = nonceGenerator.setNonceForSession(null);
+        verify(mockSession).getData();
+        assertEquals(NONCE, retNonce);
+    }
+
+    @Test
+    @DisplayName("Generate Nonce creates unique values, brute force test")
+    void generateNonceGeneratesUniqueValues() {
+        List<String> nonces = new ArrayList<>();
```
ch-dclarke, line 69: For this test, it might be best to instantiate a real `NonceGenerator` so that mocking doesn't interfere.

src/main/java/uk/gov/ch/oauth/Oauth2.java
CH-JonathanBlake: Possibly.
I think we may want a more general discussion on how we're expecting this library to be integrated into applications as our autowiring requires everywhere to setup config for those beans.

src/test/java/uk/gov/ch/oauth/nonce/NonceGeneratorTest.java
CH-JonathanBlake: Its spied and there is no setup method. There should be no mocking.

src/test/java/uk/gov/ch/oauth/nonce/NonceGeneratorTest.java
CH-JonathanBlake: Removed and noted.
Test performance improved from 1.3-1.7 seconds to 1.1-1.5 seconds.
Changing it to just check the size of the set reduced it to 1.08 - 1.11

src/test/java/uk/gov/ch/oauth/nonce/NonceGeneratorTest.java
```diff
@@ -66,12 +65,11 @@ void doesNotErrorWhenNonceIsNull() {
...
     void generateNonceGeneratesUniqueValues() {
-        List<String> nonces = new ArrayList<>();
-        for (int i = 0; i < 100; i++) {
+        Set<String> nonces = new TreeSet<>();
+        final int iterations = 100;
+        for (int i = 0; i < iterations; i++) {
             String newNonce = nonceGenerator.generateNonce();
-            System.out.println(newNonce);
-            assertFalse(nonces.contains(newNonce));
-            nonces.add(newNonce);
```
ch-dclarke: I don't see it being added to the Set...
I should have also been more explicit in my prior comment, you can do `assertTrue(nonces.add(newNonce));` as you go along.
The performance improvement that you mentioned will be more significant the larger the number of elements added - maybe you could up `iterations` to 1000

> public boolean add(E e)
> Adds the specified element to this set if it is not already present. More formally, adds the specified element e to this set if the set contains no element e2 such that (e==null ? e2==null : e.equals(e2)). If this set already contains the element, the call leaves the set unchanged and returns false.
>

CH-JonathanBlake: I don't know why this was raised against an outdated version of the code.
I realised that I could do assertTrue however in order to save those precious milliseconds I dropped the check.

src/main/java/uk/gov/ch/oauth/session/SessionUtils.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    Session getSessionByCookieId(final String cookieId) {
+        return uk.gov.companieshouse.session.SessionFactory.getSessionByCookieId(cookieId);
+    }
+
+    public Session createSession() {
+        return getSessionByCookieId(null);
+    }
+}
```
Zain-Abbas1, line 31: What is the difference between this and the SessionFactory we have in Docs? Can we look to combine the classes.

Zain-Abbas1: uk.gov.ch.developer.docs.session.SessionFactory

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -118,10 +129,10 @@ public boolean oauth2VerifyNonce(final String nonce, final String sessionNonce)
      *
      * @return The Nonce String from within the session or null if not found.
      */
-    public String getSessionNonce(final Session chSession) {
+    private String getSessionNonce() {
         String oauth2Nonce = null;
         try {
-            final Map<String, Object> data = chSession.getData();
+            final Map<String, Object> data = sessionUtils.getSessionDataFromContext();
```
Zain-Abbas1, line 135: This was changed as the session fails to update when getting the session this way. Have we tested to see whether the nonce is removed from the session after signing in through to signing out.

src/main/java/uk/gov/ch/developer/docs/config/OauthBean.java
```diff
@@ -1,27 +1,29 @@
...
     @Bean
-    IIdentityProvider identityProvider() {
+    IIdentityProvider identityProvider(EnvironmentReader environmentReader) {
         return new IdentityProvider(environmentReader);
     }
+
+    @Bean
+    SessionUtils sessionUtils() {
+        return new SessionUtils();
+    }
```
Zain-Abbas1, line 28: We have a session factory that is the same as this session utils class, we need to have one or the other.

src/main/java/uk/gov/ch/oauth/session/SessionUtils.java
CH-JonathanBlake: The difference is that this is part of the oauth package which will hopefully make it easier to extract the library.

src/main/java/uk/gov/ch/developer/docs/config/OauthBean.java
CH-JonathanBlake: As I said above, Session Factory is part of the docs developer code base. I have tried to seperate these functions so that OAuth is not dependent on docs developer.

src/main/java/uk/gov/ch/oauth/Oauth2.java
CH-JonathanBlake: As we showed via inspection. This does infact work.

src/main/java/uk/gov/ch/oauth/nonce/NonceGenerator.java
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Generates a nonce, adds it to the session and then returns the nonce.
+     *
+     * @param session takes a session and sets the unique key in the data. If the session is null,
+     * generates a new one.
+     * @return Base64 encoded unique key
+     */
+    @Override
+    public String setNonceForSession(Session session) {
```
ch-dclarke, line 28: Might be better names `addNonceToSession`

CH-JonathanBlake: Fine.

PR #71, opened 2020-04-23
Feature/fa 684 callback unit tests

ch-dclarke:
Init tests for Call Back Controller and refactoring of OAuth for simplicity

__JIRA Ticket Number__
FA-684

### Type of change

* [x] Bug fix
* [ ] New feature
* [] Breaking change

### Pull request checklist


src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -142,29 +136,29 @@ private String getSessionNonce() {
...
+        UserProfileResponse userProfile = requestUserProfile(oauthToken);
+        if ((userProfile.getId() == null) || userProfile.getId().isEmpty()) {
+            return null;
+        }
         final Map<String, Object> signInData = oauthToken.saveAccessToken();
         userProfile.addUserProfileToMap(signInData);
         signInData.put(SessionKeys.SIGNED_IN.getKey(), 1);
         final String signInInfoKey = SessionKeys.SIGN_IN_INFO.getKey();
-        final Map<String, Object> sData = chSession.getData();
+        final Map<String, Object> sData = sessionFactory.getSessionDataFromContext();
```
Zain-Abbas1, line 150: No need to get the session from context again when we already have the session in scope here, please undo.

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -63,51 +58,50 @@ public Oauth2(final IIdentityProvider identityProvider, SessionFactory sessionFa
...
-            LOGGER.error("Could not decode OAuth state", e);
-            payload = null;
+    @Override
+    public boolean isValid(final String state, final String code) {
+        final String returnedNonce = getNonceFromState(state);
+        boolean valid = oauth2VerifyNonce(returnedNonce);
+        if (!valid) {
+            LOGGER.error("Invalid nonce value in state");
+        } else {
+            valid = extractUserProfile(code);
```
Zain-Abbas1, line 88: I would prefer having a new boolean here as currently the default value of `valid` is true.

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -142,29 +136,29 @@ private String getSessionNonce() {
...
-        final WebClient webClient = WebClient.create();
-        final URI profileUrl = URI.create(identityProvider.getProfileUrl());
-
-        final UserProfileResponse userProfile =
-                getUserProfileResponse(oauthToken, webClient, profileUrl);
+    private UserProfileResponse fetchUserProfile(final String code) {
+        final OAuthToken oauthToken = requestOAuthToken(code);
 
+        UserProfileResponse userProfile = requestUserProfile(oauthToken);
+        if ((userProfile.getId() == null) || userProfile.getId().isEmpty()) {
```
Zain-Abbas1, line 143: Not sure this is the right way to go about implementing error handling for the userprofile. I think we need to look at implementing error handing directly on the webclient. Can we undo this please as this doesn't satisfy FA-706.

src/main/java/uk/gov/ch/oauth/Oauth2.java
CH-JonathanBlake: session was removed from the method signature, so its no longer in scope.

Zain-Abbas1: True, however we are getting the same session when verifying the nonce and then extracting the user profile. Maybe look to set it once.

src/test/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackControllerTest.java
```diff
@@ -1,49 +1,78 @@
...
     @Mock
-    private Payload payload;
-    @Mock
-    private JSONObject jsonObject;
-    @Mock
-    private IIdentityProvider identityProvider;
+    private Oauth2 oauth2;
 
     @InjectMocks
+    @Spy
```
Zain-Abbas1, line 33: We probably don't need this, doesn't look like we are monitoring interactions on this object.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -31,36 +26,21 @@
...
+        try {
+            final boolean valid = oauth.isValid(state, code);
+            if (valid) {
+                httpServletResponse.sendRedirect(identityProvider.getRedirectUriPage());
+            } else {
+                httpServletResponse.sendError(HttpServletResponse.SC_FORBIDDEN);
+            }
+        } catch (final Exception e) {
+            LOGGER.error(e);
+            httpServletResponse.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
```
Zain-Abbas1, line 42: Resolved based upon `sendError` which is a better way of dealing with the 403 exceptions, however for a 500 we must remember to deal with this in the eventual global exception handler.

src/main/java/uk/gov/ch/oauth/IOauth.java
```diff
@@ -10,19 +9,13 @@ String oauth2EncodeState(final String returnUri,
...
     Payload oauth2DecodeState(final String state);
 
-    /**
-     * Verify if a supplied Nonce corresponds with a session or internal value
-     *
-     * @param Nonce Supplied value
-     * @return true if Nonces match and are not both null
-     */
-    boolean oauth2VerifyNonce(final String Nonce);
+    //  UserProfileResponse fetchUserProfile(String code);
```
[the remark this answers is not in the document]
ch-dclarke: Redundant comment removed

src/main/java/uk/gov/ch/oauth/Oauth2.java
ch-dclarke: I understand the reasoning, I've simplified the code to do reduce the conditional nesting.

src/main/java/uk/gov/ch/oauth/Oauth2.java
ch-dclarke: The blocking webclient call `userProfileResponse.block(timeoutDuration)` made in `requestUserProfile` will return an empty UserProfile (o throw an exception if it fails). The original code that was refactored without change, checked to see if `userProfile` as `null`. Intellij pointed out that this could never occur, so this is a way that will deal with that problem.

src/main/java/uk/gov/ch/oauth/Oauth2.java
ch-dclarke: Doing it once within a method would make sense, however we are reducing the coupling of the methods by doing it this way.

src/test/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackControllerTest.java
ch-dclarke: Agreed, I have removed the `@Spy`

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
```diff
@@ -31,36 +26,21 @@
     private IOauth oauth;
 
     @GetMapping
-    public String getCallback(@RequestParam("state") String state,
-            @RequestParam("code") String code, final HttpServletRequest httpServletRequest) {
+    public void getCallback(@RequestParam("state") String state, @RequestParam("code") String code,
+            final HttpServletResponse httpServletResponse) {
         LOGGER.trace("Code:" + code);
         LOGGER.trace("State:" + state);
```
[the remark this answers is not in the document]
Zain-Abbas1: Can be removed, don't need them anymore.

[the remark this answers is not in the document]
ch-dclarke: Left over trace statements removed.

src/main/java/uk/gov/ch/developer/docs/controller/developer/UserCallbackController.java
CH-JonathanBlake: Set status is used because send error can throw IOExceptions which we have to add to the signature.
If we do want a global controller advice to handle this we should replace this with a throw of a custom exception.

src/main/java/uk/gov/ch/oauth/Oauth2.java
CH-JonathanBlake: To extend on this, if we make a call where there is no matching response we might expect to receive an empty JSON object. However Jackson might deserialise that as a real object with null values in all fields. This check is explicitly to handle that scenario.

Zain-Abbas1: I agree this provides a useful safety net, however I still think we would need to implement checks directly on the web client for error statuses such as access denied & failure. The perl impl handles these and then redirect so I still think some work needs to be done, which of course can be picked as part of it's own ticket.

ch-dclarke: They are trapped in the controller, as failures, including timeouts generate exceptions, but as an async web client is used, this will only occur during the `block` statement.

PR #73, opened 2020-04-28
Feature/fa 732 global errors

CH-JonathanBlake:

Created custom Exception Handler and new error pages.
This was branched from master as most work is done in new files and will merge back there before additional changes are made if needed to merge with CI branch.

__JIRA Ticket Number__
https://companieshouse.atlassian.net/browse/FA-732
### Type of change

* [ ] Bug fix
* [X] New feature
* [X] Breaking change

### Pull request checklist

### DIFF
--- a/src/main/java/uk/gov/ch/developer/docs/controller/BaseController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/BaseController.java
@@ -11,6 +11,7 @@ public abstract class BaseController {
     protected static final Logger LOGGER = LoggerFactory
             .getLogger(BaseController.class.getName());
 
+    //ToDo: Can this be removed?
     protected static final String ERROR_VIEW = "error";
 
     BaseController() {
@@ -22,5 +23,10 @@ public abstract class BaseController {
 
     @ModelAttribute("templateTitle")
     public abstract String getTemplateTitle();
+
+    @ModelAttribute("isErrorPage")
+    public boolean isErrorPage() {
+        return false;
+    }
 
 }
--- a/src/main/java/uk/gov/ch/developer/docs/controller/BaseController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/BaseController.java
@@ -11,9 +11,6 @@ public abstract class BaseController {
     protected static final Logger LOGGER = LoggerFactory
             .getLogger(BaseController.class.getName());
 
-    //ToDo: Can this be removed?
-    protected static final String ERROR_VIEW = "error";
-
     BaseController() {
 
     }
--- a/src/main/java/uk/gov/ch/developer/docs/controller/errors/DefaultErrorController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/errors/DefaultErrorController.java
@@ -1,0 +1,90 @@
+package uk.gov.ch.developer.docs.controller.errors;
+
+import java.util.HashMap;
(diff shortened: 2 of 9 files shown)


src/main/java/uk/gov/ch/developer/docs/controller/BaseController.java
```diff
@@ -11,6 +11,7 @@
     protected static final Logger LOGGER = LoggerFactory
             .getLogger(BaseController.class.getName());
 
+    //ToDo: Can this be removed?
     protected static final String ERROR_VIEW = "error";
```
ch-dclarke, line 16: Does removing it break the code?

src/test/java/uk/gov/ch/developer/docs/controller/errors/DefaultErrorControllerTest.java
```diff
@@ -0,0 +1,62 @@
...
+    HttpServletRequest mockRequest;
+    @InjectMocks
+    @Spy
+    private DefaultErrorController controller;
+
+    @Test
+    @DisplayName("404 Error")
+    void test_GetInvalidURL_ReturnsView() {
+        when(mockRequest.getAttribute(RequestDispatcher.ERROR_STATUS_CODE)).thenReturn(404);
+        when(controller.getNotFoundPath()).thenReturn("dev-hub/error/404error");
```
ch-dclarke, line 35: Make this `thenReturn(NOT_FOUND_VIEW)` instead of a hard coded value.

src/test/java/uk/gov/ch/developer/docs/controller/errors/DefaultErrorControllerTest.java
```diff
@@ -0,0 +1,62 @@
...
+        String view = controller.get();
+
+        assertEquals(NOT_FOUND_VIEW, view);
+    }
+
+    @Test
+    @DisplayName("Server Error")
+    void test_InternalServerError_ReturnsView() {
+        when(mockRequest.getAttribute(RequestDispatcher.ERROR_STATUS_CODE)).thenReturn(500);
+        when(controller.getServiceErrorPath()).thenReturn("dev-hub/error/serviceError");
```
ch-dclarke, line 46: Use your constants as return results.

src/main/java/uk/gov/ch/developer/docs/controller/BaseController.java
CH-JonathanBlake: That was more a question "Why was this added in the first place because its not used."

src/main/java/uk/gov/ch/developer/docs/controller/errors/DefaultErrorController.java
```diff
@@ -0,0 +1,102 @@
...
+        return true;
+    }
+
+    private String handleError() {
+        // get error status
+        Object status = request.getAttribute(RequestDispatcher.ERROR_STATUS_CODE);
+
+        logError();
+
+        if (status != null) {
```
Ngriffiths1, line 61: Not a request to change but i'm not a huge fan of nested statements

CH-JonathanBlake: Agreed. 
However this needs to be done before we can safely check statusCode without throwing null pointers and if we extended the nested checks with a not null check that would duplicate code and make it more prone to mistakes if we extend this.

PR #76, opened 2020-06-03
Feature/user oauth/refactor signout controller

ch-rfishwick:
Refactored Signout Controller, lifting most of the logic to Oauth2 class, and refactored unit tests accordingly. 

__JIRA Ticket Number__
FA-739

### Type of change

* [ ] Bug fix
* [ ] New feature
* [X] Breaking change

### Pull request checklist

### DIFF
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
@@ -29,9 +29,6 @@ public class SignOutController {
     @Autowired
     private Oauth2 oauth2;
 
-    public SignOutController() {
-    }
-
     @GetMapping("${signout.url}")
     public void doSignOut(final HttpServletResponse httpServletResponse,
                           final HttpServletRequest httpServletRequest) throws IOException {
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
@@ -1,7 +1,6 @@
 package uk.gov.ch.developer.docs.controller.developer;
 
 import java.io.IOException;
-import java.util.Map;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -30,9 +29,12 @@ public class SignOutController {
     @Autowired
     private Oauth2 oauth2;
 
+    public SignOutController() {
+    }
+
     @GetMapping("${signout.url}")
     public void doSignOut(final HttpServletResponse httpServletResponse,
-            final HttpServletRequest httpServletRequest) throws IOException {
+                          final HttpServletRequest httpServletRequest) throws IOException {
 
         LOGGER.debug("Processing sign out");
 
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
@@ -9,7 +9,6 @@ import uk.gov.ch.oauth.identity.IIdentityProvider;
 import uk.gov.companieshouse.logging.Logger;
 import uk.gov.companieshouse.logging.LoggerFactory;
 import uk.gov.companieshouse.session.Session;
-import uk.gov.companieshouse.session.SessionKeys;
 import uk.gov.companieshouse.session.handler.SessionHandler;
 import uk.gov.companieshouse.session.store.Store;
 
@@ -21,7 +20,7 @@ import java.io.IOException;
 public class SignOutController {
     private static final Logger LOGGER =
             LoggerFactory.getLogger(DocsWebApplication.APPLICATION_NAME_SPACE);
-    private static final String SIGN_IN_INFO = SessionKeys.SIGN_IN_INFO.getKey();
+    //private static final String SIGN_IN_INFO = SessionKeys.SIGN_IN_INFO.getKey();
 
     @Autowired
     private IIdentityProvider identityProviders;
@@ -41,7 +40,7 @@ public class SignOutController {
 
         final Session chSession = (Session) httpServletRequest
                 .getAttribute(SessionHandler.CHS_SESSION_REQUEST_ATT_KEY);
-        oauth2.invalidateSession(chSession, SIGN_IN_INFO, store);
+        oauth2.invalidateSession(chSession, store);
         httpServletResponse.sendRedirect(identityProviders.getRedirectUriPage());
     }
 
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
@@ -8,6 +8,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.GetMapping;
 import uk.gov.ch.developer.docs.DocsWebApplication;
+import uk.gov.ch.oauth.Oauth2;
 import uk.gov.ch.oauth.identity.IIdentityProvider;
 import uk.gov.companieshouse.logging.Logger;
 import uk.gov.companieshouse.logging.LoggerFactory;
@@ -26,6 +27,8 @@ public class SignOutController {
     private IIdentityProvider identityProviders;
     @Autowired
     private Store store;
+    @Autowired
+    private Oauth2 oauth2;
 
     @GetMapping("${signout.url}")
     public void doSignOut(final HttpServletResponse httpServletResponse,
@@ -36,22 +39,7 @@ public class SignOutController {
         final Session chSession = (Session) httpServletRequest
                 .getAttribute(SessionHandler.CHS_SESSION_REQUEST_ATT_KEY);
 
-        if (chSession.getSignInInfo().isSignedIn()) {
-
-            final Map<String, Object> sessionData = chSession.getData();
-            final Map<String, Object> signInInfo =
-                    (Map<String, Object>) sessionData.get(SIGN_IN_INFO);
-
-            signInInfo.replace(SessionKeys.SIGNED_IN.getKey(), 1, 0);
-            sessionData.remove(SIGN_IN_INFO);
-
-            final String zxsKey = (String) sessionData.get(".zxs_key");// This is the id of cookie
-                                                                       // stored in redis
-            if (zxsKey != null) {
-                LOGGER.trace("Deleting ZXS info from cache");
-                store.delete(zxsKey);
-            }
-        }
+        oauth2.invalidateSession(chSession, SIGN_IN_INFO, store);
         httpServletResponse.sendRedirect(identityProviders.getRedirectUriPage());
     }
 
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
@@ -7,7 +7,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.GetMapping;
 import uk.gov.ch.developer.docs.DocsWebApplication;
-import uk.gov.ch.oauth.Oauth2;
+import uk.gov.ch.oauth.IOauth;
 import uk.gov.ch.oauth.identity.IIdentityProvider;
 import uk.gov.companieshouse.logging.Logger;
 import uk.gov.companieshouse.logging.LoggerFactory;
@@ -22,7 +22,7 @@ public class SignOutController {
     @Autowired
     private IIdentityProvider identityProviders;
     @Autowired
-    private Oauth2 oauth2;
+    private IOauth oauth2;
 
     @GetMapping("${signout.url}")
     public void doSignOut(final HttpServletResponse httpServletResponse,
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
@@ -13,7 +13,6 @@ import uk.gov.companieshouse.logging.Logger;
 import uk.gov.companieshouse.logging.LoggerFactory;
 import uk.gov.companieshouse.session.Session;
 import uk.gov.companieshouse.session.handler.SessionHandler;
-import uk.gov.companieshouse.session.store.Store;
 
 @Controller
 public class SignOutController {
@@ -22,8 +21,6 @@ public class SignOutController {
 
     @Autowired
     private IIdentityProvider identityProviders;
-    @Autowired
-    private Store store;
     @Autowired
     private Oauth2 oauth2;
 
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
```diff
@@ -26,32 +27,21 @@
...
-            sessionData.remove(SIGN_IN_INFO);
-
-            final String zxsKey = (String) sessionData.get(".zxs_key");// This is the id of cookie
-                                                                       // stored in redis
-            if (zxsKey != null) {
-                LOGGER.trace("Deleting ZXS info from cache");
-                store.delete(zxsKey);
-            }
-        }
+        oauth2.invalidateSession(chSession, SIGN_IN_INFO, store);
```
CH-JonathanBlake, line 44: Why not take SIGN_IN_INFO and store into the oauth2 stuff so that we don't need to reference them and setup the store bean?

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -201,4 +202,26 @@ private void regenerateSessionID(HttpServletResponse httpServletResponse) {
...
         httpServletResponse.addCookie(buildSessionCookie(session));
     }
+
+    public void invalidateSession(Session chSession, String signInfo, Store store){
+        final Map<String, Object> sessionData = chSession.getData();
+        if (chSession.getSignInInfo().isSignedIn()){
+            removeSessionInfo(sessionData, signInfo);
+            removeZXSInfo(sessionData, store);
+        }
+    }
```
CH-JonathanBlake, line 212: If this is public could you setup Javadocs.

src/test/java/uk/gov/ch/developer/docs/controller/developer/SignOutControllerTest.java
```diff
@@ -39,87 +34,21 @@
     private HttpServletResponse httpServletResponse;
     @Mock
     private Session session;
+    @Mock
+    private Oauth2 oauth2;
```
CH-JonathanBlake, line 38: Why import this if its never used?

src/test/java/uk/gov/ch/oauth/Oauth2Test.java
```diff
@@ -0,0 +1,83 @@
...
+        when(session.getData()).thenReturn(data);
+        oauth2.invalidateSession(session, SessionKeys.SIGN_IN_INFO.getKey(), store);
+
+        assertFalse(data.containsKey(SessionKeys.SIGN_IN_INFO.getKey()));
+        verify(store, only()).delete(zxsValue);
+
+    }
+
+    @Test
+    @DisplayName("Test that a not signed in user is unable to alter session state")
```
CH-JonathanBlake, line 49: This is a test that a user cannot be signed out twice, not that they cannot change the state in other ways.

src/test/java/uk/gov/ch/developer/docs/controller/developer/SignOutControllerTest.java
ICraig-CH: We mocked Oauth2 so that when doSignOut is run during the test, the methods in Oauth2 are mocked which allows the test to run without error. I tried running the test without the mock, to see if it was needed but the results came back with a null pointer.

src/test/java/uk/gov/ch/developer/docs/controller/developer/SignOutControllerTest.java
```diff
@@ -39,87 +34,21 @@
...
 
     @Test
-    @DisplayName("Test that a valid signed in user's session state is correctly altered")
-    public void testDoSignOutWhenSignedIn() throws IOException {
-        final String zxsValue = "0000000001z";
-
-        signInInfo.setSignedIn(true);
-
+    @DisplayName("Test that a user is redirected upon signing out")
+    public void testDoSignOut() throws IOException {
```
CH-JonathanBlake, line 47: I think we ought to verify that the invalidate method it called.

src/test/java/uk/gov/ch/developer/docs/controller/developer/SignOutControllerTest.java
ch-rfishwick: Yeah, I just tried that as well. It's not stubbed explicitly because it doesn't return anything as part of the methods normal operation, we just need a mock in there so we can test the behaviour that comes afterwards

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -201,4 +202,26 @@ private void regenerateSessionID(HttpServletResponse httpServletResponse) {
...
+    private void removeSessionInfo(Map<String, Object> sessionData, String signInfo) {
+            final Map<String, Object> signInInfo =
+                    (Map<String, Object>) sessionData.get(signInfo);
+            signInInfo.replace(SessionKeys.SIGNED_IN.getKey(), 1, 0);
+            sessionData.remove(signInfo);
+    }
+    private void removeZXSInfo(Map<String, Object> sessionData, Store store) {
+        final String zxsKey = (String) sessionData.get(".zxs_key");// This is the id of cookie
+        // stored in redis
```
CH-JonathanBlake, line 221: Could we change this to a block comment or align them so its more clear that stored in redis  doesn't refer to the next line.

src/test/java/uk/gov/ch/developer/docs/controller/developer/SignOutControllerTest.java
CH-JonathanBlake: Does it matter if it always returns null, should we be testing that only the correct methods are invoked?

ICraig-CH: It can not reach the redirect if the mock is not there, because of the null pointer and for the test we are verifying if the sendRedirect is run

src/test/java/uk/gov/ch/developer/docs/controller/developer/SignOutControllerTest.java
ICraig-CH: I have added a new test above the testDoSignOut that verifies that invalidateSession is called.

src/main/java/uk/gov/ch/oauth/Oauth2.java
ICraig-CH: I will move the comment onto the one line so that it is clear what line it references.

src/test/java/uk/gov/ch/developer/docs/controller/developer/SignOutControllerTest.java
CH-JonathanBlake: Great that will resolve my concerns above I assume.

src/main/java/uk/gov/ch/oauth/Oauth2.java
ch-rfishwick: Ok, I may need you to remind me how because I have forgotten and the interwebs is not being very helpful

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
ch-rfishwick: Done

ch-rfishwick: Correction. The SIGN_IN_INFO has been moved. It was not happy about moving the store

src/test/java/uk/gov/ch/oauth/Oauth2Test.java
ch-rfishwick: True. The name and most of the logic of this test was extracted from a previous signOutController test which similarly only tested that the user couldn't sign out if they were already signed out, rather than whether or not they could change the entire state, so I have changed the name of the test to reflect that

ch-rfishwick: Checking that a user who is not signed in cannot change other aspects of the state seems more suited to a test of the sign in process right?

src/main/java/uk/gov/ch/oauth/Oauth2.java
ch-rfishwick: Changed

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
```diff
@@ -1,57 +1,44 @@
...
-    private static final String SIGN_IN_INFO = SessionKeys.SIGN_IN_INFO.getKey();
 
     @Autowired
     private IIdentityProvider identityProviders;
     @Autowired
     private Store store;
+    @Autowired
+    private Oauth2 oauth2;
+
+    public SignOutController() {
```
Zain-Abbas1, line 30: Why have we got this empty constructor?

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -201,4 +205,26 @@ private void regenerateSessionID(HttpServletResponse httpServletResponse) {
...
+        final Map<String, Object> sessionData = chSession.getData();
+        if (chSession.getSignInInfo().isSignedIn()) {
+            removeSessionInfo(sessionData, SIGN_IN_INFO);
+            removeZXSInfo(sessionData, store);
+        }
+    }
+
+    private void removeSessionInfo(Map<String, Object> sessionData, String signInfo) {
+        final Map<String, Object> signInInfo =
+                (Map<String, Object>) sessionData.get(signInfo);
```
Zain-Abbas1, line 219: Don't need `signInfo` parameter if SIGN_IN_INFO is a constant. We can call the constant directly instead of passing into method.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
```diff
@@ -1,57 +1,44 @@
...
 @Controller
 public class SignOutController {
     private static final Logger LOGGER =
             LoggerFactory.getLogger(DocsWebApplication.APPLICATION_NAME_SPACE);
-    private static final String SIGN_IN_INFO = SessionKeys.SIGN_IN_INFO.getKey();
 
     @Autowired
     private IIdentityProvider identityProviders;
     @Autowired
     private Store store;
```
Zain-Abbas1, line 26: I must say that I don't particularly like accessing the store and manipulating directly in the controller. Could there be another way of doing this, I think an obvious way could be to move it into the Oauth2 class as we're only using it there.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
```diff
@@ -1,57 +1,44 @@
...
-            sessionData.remove(SIGN_IN_INFO);
-
-            final String zxsKey = (String) sessionData.get(".zxs_key");// This is the id of cookie
-                                                                       // stored in redis
-            if (zxsKey != null) {
-                LOGGER.trace("Deleting ZXS info from cache");
-                store.delete(zxsKey);
-            }
-        }
+        oauth2.invalidateSession(chSession, store);
```
CH-JonathanBlake, line 41: Okay discussed this with Zain.
I cannot think of any scenario where the store used wouldn't be the default impl because its implicitly used in the filters and session processes, so a custom would also need to overwrite all of that as well.
Additionally, I'd be suspicious of a non reddis store because I don't think other CH apps could communicate with it. And any reasons for changing while maintaining the reddis store such as bug fixing should be done at the default level.
Therefore we should use SessionFactory.getDefaultStoreImpl() rather than requiring the app to provide the store.
This would actually be a better implementation than our bean config because it means its not our responsibility to know what the default store is.

CH-JonathanBlake: It may also be better to pass the request so that we don't rely on libraries that implement this don't have to extract their own session and incase we need additional attributes in the future we don't need to change the signature. 
But haven't discussed that yet.

src/main/java/uk/gov/ch/oauth/Oauth2.java
ICraig-CH: I have altered the code so that the constant is being directly called.

src/main/java/uk/gov/ch/oauth/Oauth2.java
CH-JonathanBlake: They look like this 
/**
  *
  */
IntelliJ is good at putting the things inbetween so just do the first bit then hit enter

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
CH-JonathanBlake: What was the problem?

CH-JonathanBlake: Did you try to move it as a spring bean?

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
ICraig-CH: I removed the constructor and then ran the tests and they still passed so I committed the change.

src/test/java/uk/gov/ch/oauth/Oauth2Test.java
CH-JonathanBlake: Honestly? It just seems like a wrong test. Users who are signed in should be able to change state.
What we want is that the only changes to state are the ones we expect in any given method.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
Zain-Abbas1: In response to Jon's first comment I agree, however we should implement `SessionFactory.getDefaultStoreImpl()` in our local uk.gov.ch.oauth.session.SessionFactory and have a method that calls off to the `SessionFactory.getDefaultStoreImpl()`. This would allow us to mock when unit testing as the original method is a static one. Furthermore, we could probably do without the store config we have. I would ensure that nothing else uses the config before deleting however.

src/test/java/uk/gov/ch/oauth/Oauth2Test.java
ch-rfishwick: How would we do that beyond testing every method that is supposed to be exclusive to a signed-in user?

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
ch-rfishwick: Changed

src/main/java/uk/gov/ch/oauth/Oauth2.java
ch-rfishwick: Done

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
ch-rfishwick: Sooo....we should modify our `oauth.session.SessionFactory ` to contain `SessionFactory.getDefaultStoreImpl()` method?

ch-rfishwick: You may need to talk us through what good that actually does wrt the `store`

src/test/java/uk/gov/ch/oauth/Oauth2Test.java
CH-JonathanBlake: Erm no we're not understanding each other. I meant just testing that nothing is changing in the state in each method that is unexpected should be enought for each method.
If the apps that use this allow a signed in user to sign in again, Accounts should deal with that.
If the apps allow a signed out user to sign out that is again Accounts problem.
If the apps allow a user to make changes to their details outside of the normal process, thats probably everyones problem, but we're not able to fix it.

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignOutController.java
CH-JonathanBlake: Sure. Set up a call and ping the chat if you need us.

PR #79, opened 2020-06-08
Refactored SignInController moving the logic to Oauth2, including get…

ICraig-CH:

Refactored SignInController moving the logic to Oauth2, including getting the original request Url and preparing the state. Altered tests to match accordingly for both SignInControllerTest and Oauth2Test.

__JIRA Ticket Number__

FA-738

### Type of change

* [ ] Bug fix
* [ ] New feature
* [X] Breaking change

### Pull request checklist

### DIFF
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
@@ -1,15 +1,16 @@
 package uk.gov.ch.developer.docs.controller.developer;
 
+import java.io.IOException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import uk.gov.ch.oauth.IOauth;
 import uk.gov.ch.oauth.identity.IIdentityProvider;
-
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-import java.io.IOException;
+import uk.gov.companieshouse.session.Session;
+import uk.gov.companieshouse.session.handler.SessionHandler;
 
 @Controller
 @RequestMapping("${signin.url}")
@@ -26,7 +27,8 @@ public class SignInController {
     @GetMapping
     public void doSignIn(final HttpServletRequest httpServletRequest,
                          final HttpServletResponse httpServletResponse) throws IOException {
-        final String state = oauth2.prepareState(httpServletRequest);
+        final Session chSession = (Session) httpServletRequest.getAttribute(SessionHandler.CHS_SESSION_REQUEST_ATT_KEY);
+        final String state = oauth2.prepareState(httpServletRequest, chSession);
         final String authoriseUri = identityProvider.getAuthorisationUrl(state);
         httpServletResponse.sendRedirect(authoriseUri);
     }
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
@@ -16,7 +16,7 @@ public class SignInController {
 
 
     @Autowired
-    private IOauth oauth2;
+    private IOauth oauth;
 
     @Autowired
     IIdentityProvider identityProvider;
@@ -25,7 +25,7 @@ public class SignInController {
     @GetMapping
     public void doSignIn(final HttpServletRequest httpServletRequest,
                          final HttpServletResponse httpServletResponse) throws IOException {
-        final String state = oauth2.prepareState(httpServletRequest);
+        final String state = oauth.prepareState(httpServletRequest);
         final String authoriseUri = identityProvider.getAuthorisationUrl(state);
         httpServletResponse.sendRedirect(authoriseUri);
     }
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
@@ -9,8 +9,6 @@ import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import uk.gov.ch.oauth.IOauth;
 import uk.gov.ch.oauth.identity.IIdentityProvider;
-import uk.gov.companieshouse.session.Session;
-import uk.gov.companieshouse.session.handler.SessionHandler;
 
 @Controller
 @RequestMapping("${signin.url}")
@@ -27,8 +25,7 @@ public class SignInController {
     @GetMapping
     public void doSignIn(final HttpServletRequest httpServletRequest,
                          final HttpServletResponse httpServletResponse) throws IOException {
-        final Session chSession = (Session) httpServletRequest.getAttribute(SessionHandler.CHS_SESSION_REQUEST_ATT_KEY);
-        final String state = oauth2.prepareState(httpServletRequest, chSession);
+        final String state = oauth2.prepareState(httpServletRequest);
         final String authoriseUri = identityProvider.getAuthorisationUrl(state);
         httpServletResponse.sendRedirect(authoriseUri);
     }
--- a/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
+++ b/src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
@@ -1,63 +1,33 @@
 package uk.gov.ch.developer.docs.controller.developer;
 
-import java.io.IOException;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/ch/oauth/Oauth2Test.java
```diff
@@ -1,55 +1,61 @@
...
     @Mock
     public IIdentityProvider identityProvider;
     @Mock
+    private OAuth2StateHandler oAuth2StateHandler;
+    @Mock
+    private NonceGenerator nonceGenerator;
+    @Mock
     public HttpServletResponse httpServletResponse;
     @Mock
     public ClientResponse response;
```
CH-JonathanBlake, line 61: These don't appear to be used and I don't think they would be injected.

src/test/java/uk/gov/ch/oauth/Oauth2Test.java
```diff
@@ -175,6 +183,27 @@ public void testSignoutWhenUserHasAnInvalidKey() {
...
+        when(request.getQueryString()).thenReturn("original");
+        String actualUrl = oauth2.getOriginalRequestURL(request);
+        assertEquals(ORIGINAL_REQUEST_URL, actualUrl);
+
+    }
+
+    @Test
+    @DisplayName("Test that prepareState returns a valid State String")
+    public void testPrepareState() {
+        doReturn(STATE).when(oauth2).encodeSignInState(anyString(), any(Session.class), anyString());
```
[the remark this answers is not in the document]
CH-JonathanBlake: Or verify that the correct arguments have been used.

src/test/java/uk/gov/ch/oauth/Oauth2Test.java
```diff
@@ -1,55 +1,61 @@
...
+
 @ExtendWith(MockitoExtension.class)
 public class Oauth2Test {
 
     public static MockWebServer mockServer;
     private static OAuthToken oauthToken;
     private final SignInInfo signInInfo = new SignInInfo();
+    static final String ORIGINAL_REQUEST_URL = "https://www.example.com?original";
+    static final StringBuffer REQUEST_URL_STRING_BUFFER = new StringBuffer(
+            "https://www.example.com");
```
CH-JonathanBlake, line 48: I assume this isn't your decision but I think its a bit strange we're expecting StringBuffers. Zain, do you know why we're doing this?

src/test/java/uk/gov/ch/oauth/Oauth2Test.java
ICraig-CH: I have removed the unused imports making sure to run the tests to make sure everything is okay.

src/test/java/uk/gov/ch/oauth/Oauth2Test.java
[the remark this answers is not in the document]
ICraig-CH: I have changed the arguments to use the available constants and a String for content. I also had to alter the tests using another doReturn on the getOriginalRequestUrl() method.

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -241,4 +245,22 @@ private void removeZXSInfo(Map<String, Object> sessionData) {
             sessionFactory.getDefaultStore().delete(zxsKey);
         }
     }
+
+    public String getOriginalRequestURL(final HttpServletRequest request) {
```
Zain-Abbas1, line 249: Can we reduce visibility of method please

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -241,4 +245,22 @@ private void removeZXSInfo(Map<String, Object> sessionData) {
...
+        final String queryString = request.getQueryString();
+        if (queryString != null) {
+            originalRequestUrl.append("?").append(queryString);
+        }
+        return originalRequestUrl.toString();
+
+    }
+
+    public String prepareState(final HttpServletRequest request) {
+        final Session chSession = (Session) request.getAttribute(SessionHandler.CHS_SESSION_REQUEST_ATT_KEY);
```
Zain-Abbas1, line 261: Can we look to get the session in the same way as the other controllers. I would check to see that functionality isn't affected by such a change however.

src/test/java/uk/gov/ch/oauth/Oauth2Test.java
Zain-Abbas1: Not entirely sure, I didn't write the initial unit test for this controller.

src/main/java/uk/gov/ch/oauth/Oauth2.java
ICraig-CH: The method is now private's and the tests refactored accordingly.

src/main/java/uk/gov/ch/oauth/Oauth2.java
ch-rfishwick: As in get the session in the actual controller, rather than in the oauth2 class?

ch-rfishwick: Changed

CH-JonathanBlake: ... But I think doing it here rather than the controller makes more sense as it gives us control incase the OAUTH changes rather than making places that implement this have to worry about maintaining it.

ch-rfishwick: 🤦

Zain-Abbas1: Yeah sorry the location was fine, I meant as opposed to getting the attribute from request and then casting to a session can we look to get the session like we've done it elsewhere. I.e, using the session factory. However, now I think about it the session factory makes a call to the Handler which essentially does the same thing by getting the servlet request attributes and casting to a session. So we could just leave it like it was, unless there is merit in the semantics or with unit testing (I can't think of any) @CH-JonathanBlake thoughts?

CH-JonathanBlake: My first thought was that we must be doing that because this is a situation where the session in the request is likely to have been altered to the store might not reflect that, this is what we do in the signout journey...
But that doesn't sound like anything we'd expect from the sign in journey.
Where would the request have added a session? Is this something that we do either in our controllers or is it actually likely to be null at this point?
This does seem like the sort of thing that we have the getSessionFromContext() method in the session factory for.

ICraig-CH: So is it fine as it is or do we need to alter parts of the SignInController?

ch-rfishwick: Ok, I can try to do that if you think that's the way to go. Currently we don't import SessionFactory into either of the controllers, is a reason why that might not be a good idea.....

CH-JonathanBlake: But we can explain why we do it in the other controller but not here.

ch-rfishwick: we don't do it in the other controller....

ch-rfishwick: Also when I try to create a sessionFactory field in the signIn controller it wants to be initialized in a constructor. Maybe we should move the session stuff to oauth2 class?

CH-JonathanBlake: Sorry perhaps I was unclear.
In the signout controller we do it this way because we are retreiving session data that has been updated but we don't know has been persisted to the store.
In the signin controller we have no reason to expect this would be case and nobody has yet said where it would even get added to the request.

ch-rfishwick: Ok, but then the point remains that I would need to add a constructor to the sign in controller...

CH-JonathanBlake: Why?
`
    public String prepareState(final HttpServletRequest request) {
        final Session chSession = sessionFactory.getSessionFromContext();
        String originalURL = getOriginalRequestURL(request);
        return encodeSignInState(originalURL, chSession, "content");
    }
`
What am I missing that is wrong with that?

ch-rfishwick: Right yeah ok, I was confused about the location still

ch-rfishwick: Changed

src/test/java/uk/gov/ch/oauth/Oauth2Test.java
ch-rfishwick: Do we want to do anything about this right now?

CH-JonathanBlake: Stick a todo on it and we can look at it during the refactor ticket?

ch-rfishwick: Done

src/main/java/uk/gov/ch/developer/docs/controller/developer/SignInController.java
```diff
@@ -9,55 +9,24 @@
...
-import uk.gov.companieshouse.session.handler.SessionHandler;
 
 @Controller
 @RequestMapping("${signin.url}")
 public class SignInController {
 
+
     @Autowired
-    IOauth oauth;
+    private IOauth oauth2;
```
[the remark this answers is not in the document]
ch-rfishwick: changed

PR #81, opened 2020-06-11
Improved testing of token classes.

CH-JonathanBlake:

Refactoring of Tokens in preparation to be moved into Oauth library

__JIRA Ticket Number__

https://companieshouse.atlassian.net/browse/FA-748

### Type of change

* [ ] Bug fix
* [ ] New feature
* [ ] Breaking change

### Pull request checklist

### DIFF
--- a/src/main/java/uk/gov/ch/oauth/Oauth2.java
+++ b/src/main/java/uk/gov/ch/oauth/Oauth2.java
@@ -1,10 +1,13 @@
 package uk.gov.ch.oauth;
 
 import static uk.gov.companieshouse.session.handler.SessionHandler.buildSessionCookie;
+
+import com.nimbusds.jose.Payload;
 import java.net.URI;
 import java.time.Duration;
 import java.util.Map;
 import javax.servlet.http.HttpServletResponse;
+import net.minidev.json.JSONObject;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.MediaType;
 import org.springframework.http.ReactiveHttpOutputMessage;
@@ -13,13 +16,12 @@ import org.springframework.web.reactive.function.BodyInserter;
 import org.springframework.web.reactive.function.BodyInserters;
 import org.springframework.web.reactive.function.client.ClientResponse;
 import org.springframework.web.reactive.function.client.WebClient;
-import com.nimbusds.jose.Payload;
-import net.minidev.json.JSONObject;
 import reactor.core.publisher.Mono;
 import uk.gov.ch.oauth.identity.IIdentityProvider;
 import uk.gov.ch.oauth.nonce.NonceGenerator;
 import uk.gov.ch.oauth.session.SessionFactory;
 import uk.gov.ch.oauth.tokens.OAuthToken;
+import uk.gov.ch.oauth.tokens.SessionSignInModifier;
 import uk.gov.ch.oauth.tokens.UserProfileResponse;
 import uk.gov.companieshouse.logging.Logger;
 import uk.gov.companieshouse.logging.LoggerFactory;
@@ -143,16 +145,16 @@ public class Oauth2 implements IOauth {
         final OAuthToken oauthToken = requestOAuthToken(code);
         regenerateSessionID(httpServletResponse);
         final UserProfileResponse userProfile = requestUserProfile(oauthToken);
+
         if ((null == userProfile.getId()) || userProfile.getId().isEmpty()) {
             return null;
         }
-        final Map<String, Object> signInData = oauthToken.saveAccessToken();
-        userProfile.addUserProfileToMap(signInData);
-        signInData.put(SessionKeys.SIGNED_IN.getKey(), 1);
-        final String signInInfoKey = SessionKeys.SIGN_IN_INFO.getKey();
-        final Map<String, Object> sData = sessionFactory.getSessionDataFromContext();
-
-        sData.merge(signInInfoKey, signInData, Oauth2::updateSignIn);
+        SessionSignInModifier sessionSignInModifier = new SessionSignInModifier();
+        sessionSignInModifier.alterSessionData(
+                sessionFactory.getSessionFromContext(),
+                oauthToken,
+                userProfile
+        );
 
         return userProfile;
     }
@@ -174,7 +176,7 @@ public class Oauth2 implements IOauth {
                 .headers(h -> h.setBearerAuth(oauthToken.getToken()))
                 .accept(MediaType.APPLICATION_JSON)
                 .exchange()
-                .flatMap(response -> validResponse(response));
+                .flatMap(this::validResponse);
         return userProfileResponse.block(timeoutDuration);
     }
 
--- a/src/main/java/uk/gov/ch/oauth/Oauth2.java
+++ b/src/main/java/uk/gov/ch/oauth/Oauth2.java
@@ -48,15 +48,6 @@ public class Oauth2 implements IOauth {
         oAuth2StateHandler = new OAuth2StateHandler(this.identityProvider);
     }
 
-    @SuppressWarnings("unchecked")
-    // This is necessary as the original data on the session is untyped, but expected to be of the correct types
-    private static Map<String, Object> updateSignIn(final Object sInf, final Object sio) {
-        final Map<String, Object> original = (Map<String, Object>) sInf;
-        final Map<String, Object> extras = (Map<String, Object>) sio;
-        original.putAll(extras);
-        return original;
-    }
-
     /**
      * {@inheritDoc}
      */
@@ -229,6 +220,7 @@ public class Oauth2 implements IOauth {
         }
     }
 
+    @SuppressWarnings("unchecked")
     private void removeSignInInfo(Map<String, Object> sessionData) {
         final Map<String, Object> signInInfo =
                 (Map<String, Object>) sessionData.get(SIGN_IN_INFO);
--- a/src/main/java/uk/gov/ch/oauth/tokens/SessionSignInModifier.java
+++ b/src/main/java/uk/gov/ch/oauth/tokens/SessionSignInModifier.java
@@ -3,6 +3,7 @@ package uk.gov.ch.oauth.tokens;
 import java.util.Map;
 import uk.gov.companieshouse.session.Session;
 import uk.gov.companieshouse.session.SessionKeys;
+import uk.gov.companieshouse.session.model.AccessToken;
 
 public class SessionSignInModifier {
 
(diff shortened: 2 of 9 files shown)


src/main/java/uk/gov/ch/oauth/tokens/SessionSignInModifier.java
```diff
@@ -0,0 +1,38 @@
...
+        SignInInfoMap signInInfo = new SignInInfoMap();
+        signInInfo.setUserProfile(userProfileResponse);
+        signInInfo.setAccessToken(accessToken);
+        signInInfo.setSignedIn(true);
+
+        alterSessionData(session.getData(), signInInfo.toMap());
+    }
+
+    public void alterSessionData(Map<String, Object> currentSessionData,
+            Map<String, Object> desiredSignInInfo) {
```
Zain-Abbas1, line 31: Does the method name have to be the same? It's confusing

src/main/java/uk/gov/ch/oauth/tokens/SessionSignInModifier.java
```diff
@@ -0,0 +1,38 @@
...
+    // This is necessary as the original data on the session is untyped, but expected to be of the correct types
+    static Map<String, Object> updateSignIn(final Object sInf, final Object sio) {
+        final Map<String, Object> original = (Map<String, Object>) sInf;
+        final Map<String, Object> extras = (Map<String, Object>) sio;
+        original.putAll(extras);
+        return original;
+    }
+
+    public void alterSessionData(Session session, AccessToken accessToken,
+            UserProfile userProfileResponse) {
```
Zain-Abbas1, line 21: Why are we passing in session, when we're not using it in this method. I think we can do without.

src/main/java/uk/gov/ch/oauth/tokens/SessionSignInModifier.java
```diff
@@ -0,0 +1,38 @@
...
+        signInInfo.setUserProfile(userProfileResponse);
+        signInInfo.setAccessToken(accessToken);
+        signInInfo.setSignedIn(true);
+
+        alterSessionData(session.getData(), signInInfo.toMap());
+    }
+
+    public void alterSessionData(Map<String, Object> currentSessionData,
+            Map<String, Object> desiredSignInInfo) {
+
```
Zain-Abbas1, line 32: Can get session data directly in this method instead of passing in `session.getData()`

src/main/java/uk/gov/ch/oauth/tokens/SessionSignInModifier.java
```diff
@@ -0,0 +1,38 @@
...
+public class SessionSignInModifier {
+
+    @SuppressWarnings("unchecked")
+    // This is necessary as the original data on the session is untyped, but expected to be of the correct types
+    static Map<String, Object> updateSignIn(final Object sInf, final Object sio) {
+        final Map<String, Object> original = (Map<String, Object>) sInf;
+        final Map<String, Object> extras = (Map<String, Object>) sio;
+        original.putAll(extras);
+        return original;
+    }
```
Zain-Abbas1, line 18: The original method from Oauth2 can be removed now

src/test/java/uk/gov/ch/oauth/tokens/OAuthTokenTest.java
```diff
@@ -14,23 +22,75 @@
...
+        String input = "{\"access_token\":\"TOKEN\",\"expires_in\":1,\"refresh_token\":\"REFRESH_TOKEN\",\"token_type\":\"TOKEN_TYPE\"}";
+        OAuthToken parsedToken = mapper.readValue(input, OAuthToken.class);
+        assertEquals(token.getTokenType(), parsedToken.getTokenType());
+        assertEquals(token.getToken(), parsedToken.getToken());
+        assertEquals(token.getRefreshToken(), parsedToken.getRefreshToken());
+        assertEquals(token.getExpiresIn(), parsedToken.getExpiresIn());
     }
+
+    private void checkNodeValue(Entry<String, JsonNode> entry, OAuthToken token) {
+        System.out.println(entry.getKey());
```
Zain-Abbas1, line 65: Delete sysout

src/test/java/uk/gov/ch/oauth/tokens/OAuthTokenTest.java
```diff
@@ -14,23 +22,75 @@
...
+    }
 
-        assertEquals(EXPIRES_IN, data.get(SessionKeys.EXPIRES_IN.getKey()));
-        assertEquals(REFRESH_TOKEN, data.get(SessionKeys.REFRESH_TOKEN.getKey()));
-        assertEquals(TOKEN, data.get(SessionKeys.ACCESS_TOKEN.getKey()));
-        assertEquals(TOKEN_TYPE, data.get(SessionKeys.TOKEN_TYPE.getKey()));
+    @Test
+    @DisplayName("Check AccessToken deserialises correctly.")
+    void setAccessTokenDeserialises() throws JsonProcessingException {
+        String input = "{\"access_token\":\"TOKEN\",\"expires_in\":1,\"refresh_token\":\"REFRESH_TOKEN\",\"token_type\":\"TOKEN_TYPE\"}";
```
Zain-Abbas1, line 56: Use `EXPIRES_IN` instead of 1

src/test/java/uk/gov/ch/oauth/tokens/SessionSignInModifierTest.java
```diff
@@ -0,0 +1,117 @@
...
+    private Map<String, Object> mockUserResponseMap;
+
+    @Test
+    void alterSessionData_withObjects_delegatesWithCorrectMaps() {
+        when(mockSession.getData()).thenReturn(sessionMap);
+        when(mockOAuthToken.getAccessTokenAsMap()).thenReturn(mockOAuthTokenMap);
+        when(mockUserResponse.getUserProfileAsMap()).thenReturn(mockUserResponseMap);
+
+        //Guys want a second opinion on whether this is good or bad practice
+        doAnswer(assertNewMapIsCorrect()).when(sessionSignInModifier)
```
Zain-Abbas1, line 50: This is confusing and hard to follow. What is the purpose of this?

src/main/java/uk/gov/ch/oauth/tokens/SessionSignInModifier.java
CH-JonathanBlake: Yes thats what overloading does. Two methods that take different arguments but give the same result.

src/main/java/uk/gov/ch/oauth/tokens/SessionSignInModifier.java
CH-JonathanBlake: I chose to pass in session here because it simplifies code on the calling side and it clearly separates the overloaded methods into ones that are interacting with the objects and ones that interact with the maps that represent data.

src/main/java/uk/gov/ch/oauth/tokens/SessionSignInModifier.java
CH-JonathanBlake: You're complaining about me passing in Session into the method that calls this. Could you please layout what you think the signatures should look like?

src/test/java/uk/gov/ch/oauth/tokens/UserProfileResponseTest.java
```diff
@@ -34,22 +41,75 @@ void setup() {
...
+        assertEquals(userProfileResponse.getLocale(), parsedToken.getLocale());
+        assertEquals(userProfileResponse.getId(), parsedToken.getId());
+        assertEquals(userProfileResponse.getForename(), parsedToken.getForename());
+        assertEquals(userProfileResponse.getSurname(), parsedToken.getSurname());
+        assertEquals(userProfileResponse.getEmail(), parsedToken.getEmail());
+    }
+
+    private void checkNodeValue(Entry<String, JsonNode> entry,
+            UserProfileResponse userProfileResponse) {
+        System.out.println(entry.getKey());
```
Zain-Abbas1, line 76: Delete sysout please

src/main/java/uk/gov/ch/oauth/tokens/SessionSignInModifier.java
CH-JonathanBlake: Good point

src/test/java/uk/gov/ch/oauth/tokens/OAuthTokenTest.java
CH-JonathanBlake: Sorry I thought I had all of them

src/test/java/uk/gov/ch/oauth/tokens/OAuthTokenTest.java
CH-JonathanBlake: you mean it should look like?
`\"expires_in\":\"EXPIRES_IN\",`?

src/test/java/uk/gov/ch/oauth/tokens/SessionSignInModifierTest.java
CH-JonathanBlake: Might be better to call.
As this method doesn't return anything and generates a new object which needs appropriate values, this catches any time that there is a call that matches this address and checks that the second argument (newMap) is correct.

src/main/java/uk/gov/ch/oauth/tokens/SessionSignInModifier.java
Zain-Abbas1: Please see my comment above, we're getting the data from request attributes, I don't see the benefit of passing in the session when we can remove it from the method signature altogether and get the session within the method. Much like how it was done prior to your change.

```suggestion
public void alterSessionData(Map<String, Object> desiredSignInInfo) {
final Map<String, Object> sData = sessionFactory.getSessionDataFromContext();
```

src/main/java/uk/gov/ch/oauth/tokens/SessionSignInModifier.java
Zain-Abbas1: I'm quite familiar with what overloading does, my point here is whether it's needed. There is no need for us to overload the method, in fact even for us to expose this method. As we would always have to set the `SignInInfoMap` data first before we go onto update the map.

src/main/java/uk/gov/ch/oauth/Oauth2.java
```diff
@@ -227,6 +220,7 @@ public void invalidateSession(Session chSession) {
         }
     }
 
+    @SuppressWarnings("unchecked")
```
ch-rfishwick, line 223: Not saying this is wrong, just curious, what is the warning here?

CH-JonathanBlake: casting the Object into a Map<String,Object>. It could throw errors if the Object isn't actually a Map.

PR #82, opened 2020-06-15
Feature/user outh/fa 751 refactor identityProvider

ICraig-CH:
Reorganised IdentityProvider class and IIdentityProvider to prepare them for extraction to a library.
Added tests for the non-getter methods of IdentityProvider.
Removed a double entry for the pom, discovered when trying to solve a problem.
__JIRA Ticket Number__
fa-751
### Type of change

* [ ] Bug fix
* [ ] New feature
* [x] Breaking change

### Pull request checklist


src/test/java/uk/gov/ch/oauth/identity/IdentityProviderTests.java
```diff
@@ -0,0 +1,52 @@
...
+@ExtendWith(MockitoExtension.class)
+public class IdentityProviderTests {
+
+
+    private IdentityProvider provider;
+
+    private EnvironmentReaderImpl reader;
+    private  String code= "code";
+    private  String state = "state";
+    private  String scope = "scope";
```
CH-JonathanBlake, line 21: Could you change these into static final constants.

ICraig-CH: Done :)

src/test/java/uk/gov/ch/oauth/identity/IdentityProviderTests.java
```diff
@@ -0,0 +1,52 @@
...
+    public void init() {
+        reader = new EnvironmentReaderImpl();
+        provider = new IdentityProvider(reader);
+    }
+
+
+    @Test
+    @DisplayName("Test that getAuthorisationUrl(final String State) returns the expected result")
+    public void testThatGetAuthorisationUrlWithStateParameterReturnsTheExpectedUrl() {
+        String expectedAuthUrl = "/oauth2/authorise?client_id=test.apps.ch.gov.test&redirect_uri=/oauth2/user/callback&response_type=code&state=state";
```
CH-JonathanBlake, line 33: I don't like the use of these magic strings but can't think of a better alternative off of the top of my head.

src/test/java/uk/gov/ch/oauth/identity/IdentityProviderTests.java
CH-JonathanBlake: Typically we write constants variable names in all caps.

src/test/java/uk/gov/ch/oauth/identity/IdentityProviderTests.java
ICraig-CH: I will look into it again to see if there is a better alternative

src/test/java/uk/gov/ch/oauth/identity/IdentityProviderTests.java
ICraig-CH: Sorry, I forgot to change that back on Friday I will change that now.

src/test/java/uk/gov/ch/oauth/identity/IdentityProviderTests.java
ICraig-CH: One solution I found was at least to make it more readable by using constants for each part of the url that is created using a getter in IdentityProvider so instead of:                          String expectedAuthUrl = "/oauth2/authorise?client_id=test.apps.ch.gov.test&redirect_uri=/oauth2/user/callback&response_type=code&state=state"        
 It would be: String expectedAuthUrl =  AUTHORISATIONURI + "?client_id=" + CLIENTID+ "&redirect_uri=" + REDIRECTURI + "&response_type=code&state=" + STATE;

CH-JonathanBlake: Isn't this getting quite close to simulating the behaviour that we're testing though?
The problem isn't that its not readable its that its fragile.
Again though. Minor problem and I don't really mind either way.

PR #93, opened 2020-07-07
Add logic to handle Signin, Callback & Signout

Zain-Abbas1:
Import oauth-java-library as dependency
Create OauthBean to initialise new coordinator and pass in application name for logging
Remove all dummy controllers
Add signin, usercallback & signout controllers to facilitate Signin & Signout
Add WebSecurity class to ensure correct filters are run on correct uri pattern
Add relevant url mappings to property file
Add all relevant unit tests

__JIRA Ticket Number__
FA-757

### Type of change

* [ ] Bug fix
* [x] New feature
* [x] Breaking change

### Pull request checklist


pom.xml
```diff
@@ -139,6 +139,35 @@
       <groupId>uk.gov.companieshouse</groupId>
       <version>${java-session-handler.version}</version>
     </dependency>
+    
+
```
ICraig-CH, line 143: Theres a few instances of blank spaces in the dependency section of the POM

PR #103, opened 2021-01-14
Fix sonar issues

Zain-Abbas1:

- Remove unused imports
- Extend class for WebSecurity

__JIRA Ticket Number__
- FA-1058

### Type of change

* [x] Bug fix
* [ ] New feature
* [ ] Breaking change

### Pull request checklist

### DIFF
--- a/src/main/java/uk/gov/ch/developer/docs/security/WebSecurity.java
+++ b/src/main/java/uk/gov/ch/developer/docs/security/WebSecurity.java
@@ -9,7 +9,7 @@ import org.springframework.security.web.authentication.www.BasicAuthenticationFi
 import uk.gov.companieshouse.session.handler.SessionHandler;
 
 @EnableWebSecurity
-public class WebSecurity {
+public class WebSecurity extends WebSecurityConfigurerAdapter {
 
     @Configuration
     @Order(1)
--- a/pom.xml
+++ b/pom.xml
@@ -42,6 +42,8 @@
     <maven.compiler.source>${java.version}</maven.compiler.source>
     <maven.compiler.target>${java.version}</maven.compiler.target>
     <maven-surefire-plugin.version>2.22.2</maven-surefire-plugin.version>
+    
+    <sonar.sources>src/main/resources</sonar.sources>
   </properties>
 
   <dependencyManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -42,8 +42,6 @@
     <maven.compiler.source>${java.version}</maven.compiler.source>
     <maven.compiler.target>${java.version}</maven.compiler.target>
     <maven-surefire-plugin.version>2.22.2</maven-surefire-plugin.version>
-    
-    <sonar.sources>src/main/resources</sonar.sources>
   </properties>
 
   <dependencyManagement>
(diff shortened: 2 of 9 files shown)


src/main/java/uk/gov/ch/developer/docs/security/WebSecurity.java
```diff
@@ -9,7 +9,7 @@
 import uk.gov.companieshouse.session.handler.SessionHandler;
 
 @EnableWebSecurity
-public class WebSecurity {
+public class WebSecurity extends WebSecurityConfigurerAdapter {
```
CH-JonathanBlake, line 12: We're creating a bean that includes inside of itself a bean of its own type?
This feels instinctively wrong to me/

Zain-Abbas1: This seems to be the only way it's done elsewhere

CH-JonathanBlake: Okay, what are the behaviour changes/ bug that got fixed?
Because from what I understood the sonar problem was just asking for a constructor to be added.

Zain-Abbas1: Yeah that issue was shown because we wasn't extending this class. Not a bug per se, but sonar issue resolved.

CH-JonathanBlake: I thought the issue said "Add an explicit constructor to hide the implicit one"?
We could have extended nearly any class to fix that.

CH-JonathanBlake: http://code-analysis.platform.aws.chdev.org:9000/component_measures?id=uk.gov.companieshouse%3Adocs.developer.ch.gov.uk&metric=new_technical_debt&selected=uk.gov.companieshouse%3Adocs.developer.ch.gov.uk%3Asrc%2Fmain%2Fjava%2Fuk%2Fgov%2Fch%2Fdeveloper%2Fdocs%2Fsecurity%2FWebSecurity.java&view=list

Add a private constructor to hide the implicit public one.
Take a look at what they offer as a compliant solution. Extending a class isn't it.

ch-rfishwick: Either solution could work for suppressing the sonar issue, as far as I can see. I understand not wanting to implicitly bring in a load of other changes without knowing exactly what all those changes are for…….but if inheriting from the security adapter is CH standard then it seems like that is a change we would have had to switch to anyway, even if we had already solved the sonar issue the way Jon is suggesting.

Do we know why it’s CH standard? Is there a reason to think that what our security class does is different in principle from what the other apps that use this inheritance are doing? If so then maybe we should consider doing it differently from them, but if not then I would say it’s better to mimic their implementation and if there are issues with then that is a problem for the organisation at large.

ICraig-CH: maybe we could bring it up with a more senior dev after the ticket is merged in ?

ch-rfishwick (approved):

ICraig-CH (approved):
I'm happy to approve based on the conversation in celestial-devs

src/main/java/uk/gov/ch/developer/docs/security/WebSecurity.java
Zain-Abbas1: After speaking with @dalebradley we've decided for consistency this change can go ahead as it is

PR #114, opened 2023-05-16
Feature/ecs application

tobafolaranmi-ch:

__JIRA Ticket Number__

### Type of change

* [ ] Bug fix
* [ ecs-docker-image] New feature
* [ ] Breaking change

### Pull request checklist

* [yes ] The code follows our [coding standards](https://github.com/companieshouse/styleguides/blob/master/java.md)


ecs-image-build/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+ARG AWS_ACCESS_KEY_ID=""
+ARG AWS_SECRET_ACCESS_KEY=""
+ARG AWS_REGION="eu-west-2"
+ARG FILE_NAME=""
+ARG PORT="8080"
+ARG JAVA_MEM_ARGS=""
+
+WORKDIR /opt
+
+RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" \
```
macreman, line 16: I think we need to find a better way of handling this. The aws cli should not be required to be installed in every production docker image and we should not have to download the zip file directly from s3 as part of this image being built.

We should be able to build this image just by copying binaries/assets from the local file system and not relying on the zip being in s3.

If we are 100% set on getting the zip from s3 for the binaries/assets (rather than copying it around within concourse) then it should be done (and unzipped) before this docker image is built

ecs-image-build/Dockerfile
```diff
@@ -0,0 +1,23 @@
+ARG IMAGE_VERSION="latest"
+FROM 416670754337.dkr.ecr.eu-west-2.amazonaws.com/ci-open-jdk8-maven-build:${IMAGE_VERSION}
```
macreman, line 2: shouldn't this be one of the "runtime" base images we've been building?

ecs-image-build/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+WORKDIR /opt
+
+RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" \
+    && unzip awscliv2.zip \
+    && ./aws/install \
+    && aws s3 cp s3://${S3_RELEASE_BUCKET}/${APPLICATION_NAME}/${FILE_NAME} . \
+    && unzip ${FILE_NAME} \
+    && rm -f ${FILE_NAME} awscliv2.zip
+
+CMD ["java",  "-jar", "-Dserver.port=8080", "docs.developer.ch.gov.uk.jar"]
```
macreman, line 23: this should be a standardised start script that we copy into the container as an asset rather than running the java command directly

macreman (changes requested):

ecs-image-build/Dockerfile
tobafolaranmi-ch: The runtime images we've recently built are jdk7 not jdk8. The application requires jdk8

macreman: Is the Java17 runtime image incompatible with the Java8 built binary? I'm pretty sure (though happy to be proven wrong!) that we can run the binary in the new java17 runtime container even if it was built with java 8

tobafolaranmi-ch: 416670754337.dkr.ecr.eu-west-2.amazonaws.com/ci-open-jdk-runtime-17 doesn't start up the application as expected.

macreman: Confirmed that the springboot version its using (2.2.0) is too old for java 17 runtime so even though it should be able to run when built from java 8 image we'd need to update the service itself to use a newer version of springboot which, from a quick google, looks like 2.5.5 https://docs.spring.io/spring-boot/docs/2.5.5/reference/html/getting-started.html#getting-started.system-requirements

This can go in with the jdk8 build container as a base image for now providing we have an item in the backlog to tackle this change asap

ecs-image-build/Dockerfile
tobafolaranmi-ch: Changes have been made in the pipeline and dockerfile to unarchive the artifact in concourse before the docker build

ecs-image-build/Dockerfile
tobafolaranmi-ch: docker_start.sh script added to start the container

docker_start.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+#
+# Start script for docs.developer.ch.gov.uk
+
+PORT=8080
+
+exec java -jar -Dserver.port="${PORT}" "docs.developer.ch.gov.uk.jar"
+#exec java ${JAVA_MEM_ARGS} -jar -Dserver.port="${PORT}" -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005 "${APP_DIR}/docs.developer.ch.gov.uk.jar"
```
sienkin, line 8: This commented command should be removed.

ecs-image-build/Dockerfile
```diff
@@ -0,0 +1,12 @@
+ARG IMAGE_VERSION="latest"
+FROM 416670754337.dkr.ecr.eu-west-2.amazonaws.com/ci-open-jdk8-maven-build:${IMAGE_VERSION}
+
+ARG APPLICATION_NAME=docs.developer.ch.gov.uk
+ARG PORT="8080"
+ARG JAVA_MEM_ARGS=""
```
sienkin, line 6: These build arguments aren't referenced further in the Dockerfile - are they required / should they be assigned to `ENV`?

ecs-image-build/docker_start.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+#
+# Start script for docs.developer.ch.gov.uk
+
+PORT=8080
+
+exec java -jar -Dserver.port="${PORT}" "docs.developer.ch.gov.uk.jar"
+#exec java ${JAVA_MEM_ARGS} -jar -Dserver.port="${PORT}" -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005 "${APP_DIR}/docs.developer.ch.gov.uk.jar"
```
sienkin, line 8: See https://github.com/companieshouse/docs.developer.ch.gov.uk/pull/114/files#r1211549049

docker_start.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+#
+# Start script for docs.developer.ch.gov.uk
+
+PORT=8080
```
sienkin, line 5: The port is hard-coded here but this is assigned as a build ARG in the Dockerfile - should the build ARG be assigned to an ENV and this updated to use the ENV instead?

ecs-image-build/docker_start.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+#
+# Start script for docs.developer.ch.gov.uk
+
+PORT=8080
```
sienkin, line 5: See https://github.com/companieshouse/docs.developer.ch.gov.uk/pull/114/files#r1211552273

docker_start.sh
tobafolaranmi-ch: ARG removed from dockerfile

ecs-image-build/docker_start.sh
tobafolaranmi-ch: ARG removed from dockerfile

sienkin (approved):

PR #121, opened 2023-11-21
CC-668-check-and-validate-log-levels

ivana-companieshouse:

__JIRA Ticket Number__

### Type of change

* [ ] Bug fix
* [ ] New feature
* [ ] Breaking change

### Pull request checklist


JMetherell (approved):
I think the log level on staging should be => live, but approving as you confirmed that is what you intended to set.

